PR #55, opened 2018-07-16
Set the top padding of two columns to be equal

ssantichaivekin:
I forgot to branch in my first PR (open links in new tab) so this PR also has the commit.

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -176,7 +176,7 @@ body {
 /** Selected courses **/
 
 #selected-courses-column {
-    padding-top: 10px;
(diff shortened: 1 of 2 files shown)


raxod502:
You don't have to base the feature branch on the latest `master`. If you based it on `origin/master` instead, then this pull request would have only the one commit you wanted on it, even though you had additional commits on your fork's `master` branch. You don't need to fix it in this case since your other pull request was merged, though.

raxod502:
Also, it's not terribly important, but generally commit message summary lines should be limited to 50 characters.

PR #57, opened 2018-07-16
Implement infinite scrolling

ssantichaivekin:
Note that scroll bar size is quite a problem. The scroll bar starts out pretty big, indicating that there is little content below even when that is a lot of content.

(Part of this PR is from another PR that I forgot to branch out while developing.)

screen shot 2018-07-16 at 12 41 58 am

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -146,33 +146,6 @@ body {
     text-align: center;
 }
 
-/*** Course search pagination bar ***/
-
-#course-search-pagination-bar {
-    padding-top: 15px;
-    padding-bottom: 20px;
-}
-
-#course-search-fewer-pages-button-wrapper {
-    display: inline-block;
-    padding-right: 10px;
-}
-
-#course-search-more-pages-button-wrapper {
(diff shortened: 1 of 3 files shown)


raxod502:
I don't think the scroll bar size is a problem. Everyone knows this is what infinite-scroll looks like.

raxod502:
This seems like it would impact performance on mobile. Are you sure it's not too slow to do this kind of update every 300ms?

ssantichaivekin:
I believe checking every 300ms won't be a problem because each check is fast if no re-layout (add more things to list) is required to run. If a re-layout is run, then it would be quite slow, but not slower than manually clicking "more page".

My understanding is that a computer can run ≈1M trivial lines of javascript code per second, so the check alone (like 10 lines per 1/3 sec?) should not affect performance (again, if no things are added).

Also, it would be nicer to have an end-of-result label on the bottom.

ssantichaivekin:
Maybe I will recheck the code before submitting a new PR.

PR #59, opened 2018-07-18
Implement infinite scroll

ssantichaivekin:
Implement infinite scroll
Implement "End of Result" text under results

Performance concern:
Checking whether we have to run "updateCourseSearchResults" every 300ms is very fast. The update itself, however, is slow because it always rerun search filter on gCourseList from the first item until we get to the number of items we want.

Note that this infinite scroll feature is not slower than manually clicking next page.

The updateCourseSearchResults re-paginate time can be reduced greatly by saving the last index visited in gCourseList. When we re-paginate, we can then start from the last index visited populate the list with more children.

### DIFF
--- a/src/js/hyperschedule.js
+++ b/src/js/hyperschedule.js
@@ -10,6 +10,9 @@ const millisecondsPerHour = 3600 * 1000;
 
 const courseSearchPageSize = 20;
 
+const courseSearchPagesShownStart = 3;
+const extraPagesToLoad = 2;
+
 const apiURL = API_URL; // replaced by Babel with a string literal
 
 //// DOM elements
@@ -22,13 +25,9 @@ const closedCoursesToggle = document.getElementById("closed-courses-toggle");
 const courseSearchColumn = document.getElementById("course-search-column");
 const scheduleColumn = document.getElementById("schedule-column");
 
+const courseSearchScheduleColumn = document.getElementById("course-search-schedule-column");
 const courseSearchInput = document.getElementById("course-search-course-name-input");
 const courseSearchResultsList = document.getElementById("course-search-results-list");
-
-const courseSearchFewerPagesButton = document.getElementById("course-search-fewer-pages-button");
-const courseSearchMorePagesButton = document.getElementById("course-search-more-pages-button");
-const courseSearchOnePageButton = document.getElementById("course-search-one-page-button");
-const courseSearchAllPagesButton = document.getElementById("course-search-all-pages-button");
 
 const importExportDataButton = document.getElementById("import-export-data-button");
 
@@ -61,7 +60,11 @@ let gShowClosedCourses = true;
 let gCourseIndex = {};
 let gSelectedCoursesIndex = {};
 let gCurrentlySorting = false;
-let gCourseSearchPagesShown = 1;
+let gCourseSearchPagesShown = courseSearchPagesShownStart;
+// gMaxCourseSearchPage starts as Infinity and stays infinity until we
+// exhaust the search result, at which point, it is calculated and set.
+let gMaxCourseSearchPage = Infinity;
+let gNextIncrementalCourseSearchIndex = null;
 
 /// Utility functions
 //// JavaScript utility functions
@@ -393,11 +396,6 @@ function courseMatchesSearchQuery(course, query)
   return true;
 }
 
-function maxCourseSearchPages(courseList)
-{
-  return Math.ceil(courseList.length / courseSearchPageSize);
-}
-
 ///// Course scheduling
 
 function generateScheduleSlotDescription(slot)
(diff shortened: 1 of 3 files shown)


ssantichaivekin:
Note that I left some unreachable pagination code in the file (fewer pages logic and things like that).

src/js/hyperschedule.js
```diff
@@ -56,12 +52,15 @@ let gCourseDataTimestamp = null;
 let gSelectedCourses = [];
 let gScheduleTabSelected = false;
 let gShowClosedCourses = true;
+let gCourseSearchPagesShownStart = 3;
+let extraPagesToLoad = 2;
```
raxod502: These look like constants. They should go in the section for constants, not globals.

src/js/hyperschedule.js
```diff
@@ -56,12 +52,15 @@ let gCourseDataTimestamp = null;
...
 
 // Transient data.
 let gCourseIndex = {};
 let gSelectedCoursesIndex = {};
 let gCurrentlySorting = false;
-let gCourseSearchPagesShown = 1;
+let gCourseSearchPagesShown = gCourseSearchPagesShownStart;
+let endOfResultDisplayed = false;
```
raxod502: Please use the `g` prefix if this is a mutable global variable.

src/js/hyperschedule.js
```diff
@@ -632,6 +642,25 @@ function createCourseLoadingMessage()
...
 
+function createCourseSearchEndOfResult(hasResult = true)
+{
+  const listItem = document.createElement("li");
+  listItem.id = "course-search-placeholder";
+  listItem.setAttribute("data-placeholder", "true");
+  let text;
+  if (hasResult)
+  {
+    text = document.createTextNode("•    End of Result   •");
```
raxod502: How about "End of results" instead?

src/js/hyperschedule.js
```diff
@@ -632,6 +642,25 @@ function createCourseLoadingMessage()
...
+  }
+  else
+  {
+    text = document.createTextNode("•    No Result   •");
```
raxod502: And analogously "No results".

src/js/hyperschedule.js
```diff
@@ -1246,13 +1277,16 @@ function setCourseSearchPagesDisplayed(action)
   if (numPages !== null)
   {
     numPages = Math.max(numPages, 1);
+    // the line below is wrong.
```
raxod502: Why don't we fix it, then?

raxod502 (changes requested):
Seems reasonable, I think. To clarify, you're saying that performance when scrolling through the already-populated section of the search results is _not_ affected by this change? If that's the case, I think it's fine.

src/js/hyperschedule.js
ssantichaivekin: Don't you like the • ?

src/js/hyperschedule.js
ssantichaivekin: This bug deserves its own PR. Fixing this requires changing how we store and display search result.

The current incremental update makes it impossible to determine the number of total things to show until all the items has been calculated. This means that we do not know how many results we have left until we have search through the whole dict. If we are on a page, say page 3, we can only know either (1) this is the last page, or (2) this is not the last page.

This wrong calculation may impact the performance of the app just a bit by causing the app to unnecessarily run *one* more incremental update (which doesn't really update anything) when you reach to the bottom.

ssantichaivekin:
Scrolling through a populated section of the search result is affected by this change only when we are scrolling near the bottom and update course list is called. Scrolling up and down already updated section is not affected by the change. Update can cause lags, but the lag time is equal to the time it takes to run update when you manually click more results.

Except when we reach the bottom, we might get one extra small lag that does not happen when manually clicking more results.

src/js/hyperschedule.js
raxod502: I don't object to it. I was just talking about the capitalization and pluralization.

src/js/hyperschedule.js
raxod502: OK. Sounds fine, then.

raxod502:
OK. Your assessment of the performance sounds fine to me.

src/js/hyperschedule.js
```diff
@@ -61,7 +60,8 @@ let gShowClosedCourses = true;
 let gCourseIndex = {};
 let gSelectedCoursesIndex = {};
 let gCurrentlySorting = false;
-let gCourseSearchPagesShown = 1;
+let gCourseSearchPagesShown = courseSearchPagesShownStart;
+let gEndOfResultDisplayed = false;
```
raxod502: Why is this global? It looks like it's only used in one function.

raxod502 (changes requested):

src/js/hyperschedule.js
ssantichaivekin: When update get unnecessarily called when we reach the bottom, it can add an extra “end of results” text at the bottom.

This is a hot-fix for that. It works by remembering (globally) whether the text has been added. Note that fixing the pagination bug will allow us to remove the “is end of result already displayed?” logic altogether.

I will think about this more and maybe add some comment + open an issue to remove this after the pagination bug is fixed.

ssantichaivekin: Actually never mind. I just thought of another way to fix this.

ssantichaivekin:
Sorry for long messy PR.

- Fixed pagination bug where where we use the total course list to calculate max page number. Now, the max page number will be infinity (meaning that we don't know what the max page number is) until it is set by updateCourseSearchResults to be the correct value when it reaches the end of result. 

- Incremental update now starts from previous index it stopped last time and continue matching queries from there. The number of string-matching queries called by scrolling to the bottom of all course decreases from 97024 (quadratic total time) to 2029 (linear total time). Note that we have 1935 items in the course list (I feel like the fix should make us reach exactly 1935, I still don't know where the extra queries comes from). This will be useful, especially in the future where we want to add more search features / make search queries more expensive.

However, I noticed that decrease in number of queries does not make the scroll go any smoother. Scrolling very fast still sometimes hits the bottom and takes a bit of time to lay out more items. I experimented and changed the line of code `setTimeout(autoUpdateNumCourseSearchPagesDisplayed, 300);` to `setTimeout(autoUpdateNumCourseSearchPagesDisplayed, 100);` and notice a much better scroll. This points out that the reason the scroll is laggy is because the refresh rate is too low. I have not tested on mobile.

raxod502 (approved):

raxod502:
I rebased and fixed a few minor style things. Thanks so much for your contribution; it's a great improvement.

PR #62, opened 2018-07-31
fix course display bug

ssantichaivekin:
Fix course display bug where courses search display the result twice. The bug is caused by the line `gNextIncrementalCourseSearchIndex = courseListIndex;` not being executed when the result reaches the end.


raxod502:
Thanks.

PR #75, opened 2018-11-08
Add Information to PDFs

### LINKED ISSUE #67: PDF export should include class location, professor
Currently, the PDF export only shows course code and name. It should have the same metadata as the iCal export. cc @kwshi

Owen-Gillespie:
This commit uses the extra space in the blocks for longer classes to display extra information. At most, the course code, course name, professors (last names), and location will be displayed. If there is not enough room for everything, there is an order of priority for the information. Currently this order is
1. Code
2. Location
3. Course name
4. Professors 

Closes #67

### DIFF
--- a/src/js/hyperschedule.js
+++ b/src/js/hyperschedule.js
@@ -355,6 +355,11 @@ function courseToString(course)
     course.totalSeats + " seats filled)";
 }
 
+function courseToInstructorLastnames(course)
+{
+  return course.faculty.map(fullName => fullName.split(",")[0]).join(",");
+}
+
 function generateCourseDescription(course)
 {
   const description = [];
@@ -733,7 +738,7 @@ function createCourseSearchEndOfResult(hasResult = true)
           (gMalformedCourseCount === 1) ? "it" : "them";
     const hintText = document.createTextNode(
       `${gMalformedCourseCount} ${coursesWere} omitted from results ` +
-      `because the registrar entered malformed data for ${them} – `);
+        `because the registrar entered malformed data for ${them} – `);
     listItem.appendChild(hintText);
     const link = document.createElement("a");
     link.appendChild(document.createTextNode("click here for more details"));
@@ -1662,21 +1667,64 @@ function downloadPDF()
           course.courseName,
           width - 12,
         );
+        const courseLocationLines = pdf.splitTextToSize(
+          slot.location,
+          width - 12,
+        );
+        const courseInstructorsLines = pdf.splitTextToSize(
+          courseToInstructorLastnames(course),
+          width - 12
+        );
+
+        // Attributes to be shown are calculated based on space available
+        // order of preference: Code, location, name, professor
+        // order of display: code, name, professor, location
+
+        const entriesByPreference = ["code", "location", "name", "instructor"];
+        const entriesByOrder = ["code", "name", "instructor", "location"];
+
+        let entryNameToText = {
+          code: courseCodeLines,
+          location: courseLocationLines,
+          name: courseNameLines,
+          instructor: courseInstructorsLines,
+        };
+
+        // Limits size of text to ensure some white space at top and bottom
+        const maxTextLength = (yEnd - yStart) - 2 * pdf.getLineHeight();
+        let numEntries = 0;
+        let totalLength = 0;
+        while (totalLength * pdf.getLineHeight() < maxTextLength &&
+               numEntries < entriesByPreference.length)
+        {
+          totalLength += entryNameToText[entriesByPreference[numEntries]].length;
+          numEntries += 1;
+        }
 
         const xText = x + width / 2;
-        const yText = (yStart + yEnd) / 2 -
-              (courseCodeLines.length + courseNameLines.length) *
-              pdf.getLineHeight() / 2 +
-              pdf.getLineHeight();
+        // Find height to start the text so that it will be centered in the block
+        let yText = (yStart + yEnd) / 2 -
+            totalLength * pdf.getLineHeight() / 2 +
+            pdf.getLineHeight();
+
         pdf.setFontStyle("bold");
         pdf.text(xText, yText, courseCodeLines, "center");
+        yText += courseCodeLines.length * pdf.getLineHeight();
         pdf.setFontStyle("normal");
-        pdf.text(
-          xText,
-          yText + courseCodeLines.length * pdf.getLineHeight(),
-          courseNameLines,
-          "center",
-        );
+
+        for (let entry of entriesByOrder)
+        {
+          if (entriesByPreference.slice(1, numEntries).includes(entry))
+          {
+            pdf.text(
+              xText,
+              yText,
+              entryNameToText[entry],
+              "center",
+            );
+            yText += entryNameToText[entry].length * pdf.getLineHeight();
+          }
+        }
       }
     }
   }


src/js/hyperschedule.js
```diff
@@ -1669,21 +1669,66 @@ function downloadPDF()
           course.courseName,
           width - 12,
         );
+        const courseLocationLines = pdf.splitTextToSize(
+          slot.location,
+          width - 12,
+        );
+        const courseInstructorsLines = pdf.splitTextToSize(
+          course.faculty.map(fullName => fullName.split(",")[0]).join(", "),
```
Owen-Gillespie: The code to make a list of last names could be pulled off into a helper function

src/js/hyperschedule.js
```diff
@@ -1669,21 +1669,66 @@ function downloadPDF()
...
 
+        // Attributes to be shown are calculated based on space available
+        // order of preference: Code, location, name, professor
+        // order of display: code, name, professor, location
+        const maxTextLength = (yEnd - yStart) - 2 * pdf.getLineHeight();
+        const toLocationLength = courseCodeLines.length +
+              courseLocationLines.length;
+        const toNameLength = toLocationLength + courseNameLines.length;
+        const toInstructorLength = toNameLength + courseInstructorsLines.length;
+        const showLocation = toLocationLength * pdf.getLineHeight() <
```
Owen-Gillespie: Edge case to allow 50 minute blocks to contain most locations. Not sure how comment heavy you like your code, but could have a comment here or just get rid of this.

src/js/hyperschedule.js
```diff
@@ -1669,21 +1669,66 @@ function downloadPDF()
...
+              maxTextLength + .5 * pdf.getLineHeight();
+        const showName = toNameLength  * pdf.getLineHeight() < maxTextLength;
+        const showInstructor = toInstructorLength  * pdf.getLineHeight() < maxTextLength;
+        const totalLength = Math.max(courseCodeLines.length,
```
Owen-Gillespie: This whole construct of computing all the different possible lengths and then checking which are too long and then getting the longest one that isn't too long feels kinda gross, but I couldn't come up with a cleaner way to do it.

src/js/hyperschedule.js
```diff
@@ -1669,21 +1669,66 @@ function downloadPDF()
...
+        // Attributes to be shown are calculated based on space available
+        // order of preference: Code, location, name, professor
+        // order of display: code, name, professor, location
+        const maxTextLength = (yEnd - yStart) - 2 * pdf.getLineHeight();
+        const toLocationLength = courseCodeLines.length +
+              courseLocationLines.length;
+        const toNameLength = toLocationLength + courseNameLines.length;
+        const toInstructorLength = toNameLength + courseInstructorsLines.length;
+        const showLocation = toLocationLength * pdf.getLineHeight() <
+              maxTextLength + .5 * pdf.getLineHeight();
```
raxod502: Just for consistency I think we should use `0.5` instead of `.5`.

src/js/hyperschedule.js
raxod502: I think that's a good idea; it can be put with the other functions that take a course object and return some string.

src/js/hyperschedule.js
```diff
@@ -1669,21 +1669,66 @@ function downloadPDF()
...
+        if(showInstructor){
+            pdf.text(
+                xText,
+                yText,
+                courseInstructorsLines,
+                "center",
+            );
+            yText += courseInstructorsLines.length * pdf.getLineHeight();
+        }
+        if(showLocation){
```
raxod502: Space between `)` and `{` here. Also after `if`.

src/js/hyperschedule.js
```diff
@@ -1669,21 +1669,66 @@ function downloadPDF()
...
+        );
+        const courseInstructorsLines = pdf.splitTextToSize(
+          course.faculty.map(fullName => fullName.split(",")[0]).join(", "),
+          width - 12
+        );
 
+        // Attributes to be shown are calculated based on space available
+        // order of preference: Code, location, name, professor
+        // order of display: code, name, professor, location
+        const maxTextLength = (yEnd - yStart) - 2 * pdf.getLineHeight();
```
raxod502, line 1694: Why are we subtracting `2 * pdf.getLineHeight()` here?

src/js/hyperschedule.js
raxod502: Why does this differ from the other cases below?

src/js/hyperschedule.js
raxod502: How about making a list `["code", "location", "name", ...]` and then iterating through it and incrementing the `totalLength` until it gets too big? You can figure out how much to increment by on each iteration by looking up in a map `{ code: courseCodeLines.length, ... }`. Then by the end of the loop you know which things to add.

src/js/hyperschedule.js
```diff
@@ -1669,21 +1669,66 @@ function downloadPDF()
...
+        const toNameLength = toLocationLength + courseNameLines.length;
+        const toInstructorLength = toNameLength + courseInstructorsLines.length;
+        const showLocation = toLocationLength * pdf.getLineHeight() <
+              maxTextLength + .5 * pdf.getLineHeight();
+        const showName = toNameLength  * pdf.getLineHeight() < maxTextLength;
+        const showInstructor = toInstructorLength  * pdf.getLineHeight() < maxTextLength;
+        const totalLength = Math.max(courseCodeLines.length,
+                                    showLocation * toLocationLength,
+                                    showName * toNameLength,
+                                    showInstructor * toInstructorLength);
```
raxod502: Okay, multiplying booleans by ints is definitely bad.

src/js/hyperschedule.js
```diff
@@ -1669,21 +1669,66 @@ function downloadPDF()
...
         const xText = x + width / 2;
-        const yText = (yStart + yEnd) / 2 -
-              (courseCodeLines.length + courseNameLines.length) *
-              pdf.getLineHeight() / 2 +
+        let yText = (yStart + yEnd) / 2 -
+              totalLength * pdf.getLineHeight() / 2 +
               pdf.getLineHeight();
```
raxod502: I think the idea behind these calculations should probably be explained a bit. What are we computing, and then why do the offsets make sense?

src/js/hyperschedule.js
raxod502: Also, in general, more comments is almost never a bad thing.

src/js/hyperschedule.js
```diff
@@ -1669,21 +1669,66 @@ function downloadPDF()
...
+        }
+        if(showLocation){
+            pdf.text(
+                xText,
+                yText,
+                courseLocationLines,
+                "center",
+            );
+            yText += courseLocationLines.length * pdf.getLineHeight();
+        }
```
raxod502, line 1727: I think this code could also be turned into a loop without that much difficulty. It would be a big improvement in my opinion.

raxod502 (changes requested):
Approach seems good to me. I think the duplicated code for each field type could be made generic by the use of maps from the field names to the content, number of lines, etc., and we should do this.

src/js/hyperschedule.js
Owen-Gillespie: Noted in comment

src/js/hyperschedule.js
Owen-Gillespie: Noted in comment

Owen-Gillespie:
@raxod502 Let me know if you want any more changes

raxod502 (approved):
Indentation is two spaces, not four. I fixed that, along with some spacing. Otherwise, it looks great. Thank you for the help and I'm sorry it took me so long to get to this.

PR #80, opened 2019-08-03
Only course list/schedule scrolls in left column

ingridt123:
In the course search, only the course list (and not the toolbar and search bar) scrolls so that the user can easily make changes to the search without having to scroll up and down. In the schedule, only the schedule (and not the toolbar) scrolls.

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -54,7 +54,6 @@ body {
 .two-column-left {
     float: left;
     height: 100%;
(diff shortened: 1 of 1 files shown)


ingridt123:
Sorry schedule doesn't scroll right now will fix and commit again

raxod502 (approved):
Wow, that makes a lot more sense :P 
Thanks for your contribution!

Issue #34, opened 2017-12-17
Show whether course is open or closed in course details pane


raxod502:
Currently, there is no way to see whether a course is open or closed without scrolling down the list of selected courses to find it. This should be changed.

ingridt123:
Which one would people prefer?
1. "(open/closed)" after the course name
2. "(open/closed, 15/16 seats filled)" after the course name
3. "Open/Closed, 15/16 seats filled" as the last entry in the course description box

<img width="372" alt="Screen Shot 2019-08-03 at 12 17 25 PM" src="https://user-images.githubusercontent.com/19776302/62416070-48820a00-b5e9-11e9-8652-4c1425a96d26.png">
<img width="371" alt="Screen Shot 2019-08-03 at 12 17 43 PM" src="https://user-images.githubusercontent.com/19776302/62416071-48820a00-b5e9-11e9-91d8-1cea4b246bf9.png">
<img width="369" alt="Screen Shot 2019-08-03 at 12 19 39 PM" src="https://user-images.githubusercontent.com/19776302/62416072-48820a00-b5e9-11e9-9761-97e8c72c22d9.png">

raxod502:
Looks like this issue is solved by Ingrid.

PR #81, opened 2019-08-05
Added enrollment status to course description box

ingridt123:
Don't have to scroll all the way to course in selected course list to view enrollment status
Issue #34


raxod502 (approved):
:sparkles: 

sorry I didn't get back to your question about design; I think you chose the best option in any case :)

PR #82, opened 2019-08-05
Put closed course toggle and help button on new rows if not enough space

ingridt123:
Otherwise course search is very small (can't see what is typed) and the help button goes on top of right column
image


raxod502 (approved):
🏢

PR #83, opened 2019-08-05
Responsive design 2

ingridt123:
also changed percentages of right and left columns so that "View on Github" would fit on laptop browser window


ingridt123:
Sorry this was an accidental pull request

PR #84, opened 2019-08-05
 Put import/export data button and print button on new rows if not enough space

ingridt123:
Also changed percentages of right/left columns so that the "View on Github" button would fit on a normal browser

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -96,7 +96,7 @@ body {
 }
 
 #schedule-column {
-    height: 93%;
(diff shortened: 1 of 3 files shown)


raxod502:
It seems that there is a conflict since I merged your first pull request. Do you think you could take a look at that?

ingridt123:
I just resolved the conflicts, not sure if I need to do anything else?

raxod502 (approved):
↕️

Issue #73, opened 2018-11-01
Gray out/highlight courses that conflict with those in the calendar


cole-k:
Probably add three options:

* None: don't gray out anything
* Starred: only gray out courses conflicting with starred courses
* All: gray out all conflicts

Similar to #20.

ingridt123:
Implemented by Pull Request #87

PR #87, opened 2019-08-12
Added option to grey out conflicted courses

ingridt123:
User can select between three options in the settings modal dialog
Screen Shot 2019-08-11 at 8 44 06 PM

Issue #73

### DIFF
--- a/src/js/hyperschedule.js
+++ b/src/js/hyperschedule.js
@@ -730,6 +730,11 @@ function attachListeners()
     gCurrentlySorting = false;
   });
 
+  // for (var i = 0; i < conflictCoursesRadios; i++) {
+  //   conflictCoursesRadios[i].addEventListener("click", () => {
+  //     gGreyConflictCourses = gGreyConflictCoursesOptions[i];
+  //   })
+  // }
   conflictCoursesRadio1.addEventListener("click", () => {
     gGreyConflictCourses = gGreyConflictCoursesOptions[0];
     toggleConflictCourses();
@@ -1507,7 +1512,7 @@ function toggleCourseStarred(course)
 {
   course.starred = !course.starred;
   updateCourseSearchResults();
-  updateSelectedCoursesList(); // TODO
+  updateSelectedCoursesList();
   updateSchedule();
   writeStateToLocalStorage();
 }
@@ -1678,7 +1683,7 @@ function writeStateToLocalStorage()
   localStorage.setItem("selectedCourses", JSON.stringify(gSelectedCourses));
   localStorage.setItem("scheduleTabSelected", gScheduleTabSelected);
   localStorage.setItem("showClosedCourses", gShowClosedCourses);
-  localStorage.setItem("greyConflictCourses", gGreyConflictCourses); //TODO
+  localStorage.setItem("greyConflictCourses", JSON.stringify(gGreyConflictCourses));
 }
 
 function oldCourseToString(course)
@@ -1757,7 +1762,7 @@ function readStateFromLocalStorage()
   );
   gGreyConflictCourses = readFromLocalStorage(
     "greyConflictCourses", _.isString, gGreyConflictCoursesOptions[1]
-  ); //TODO
+  );
 }
 
 /// PDF download
--- a/src/js/hyperschedule.js
+++ b/src/js/hyperschedule.js
@@ -1151,15 +1151,20 @@ function updateCourseSearchBar() {
 
   // default value
   let tableValue = "table-cell";
+  let marginValue = "0 auto";
 
   let minSearchInputWidth = 100;
   if (courseSearchColumn.offsetWidth < 
     (minSearchInputWidth + courseClosedToggleLabel.offsetWidth + helpButton.offsetWidth)) {
     tableValue = "table-row";
+    marginValue = "5px auto";
   }
   courseSearchInputWrapper.style.display = tableValue;
+  courseSearchInput.style.margin = marginValue;
   courseClosedToggleWrapper.style.display = tableValue;
+  courseClosedToggleLabel.style.margin = marginValue;
   helpButtonWrapper.style.display = tableValue;
+  helpButton.style.margin = marginValue;
 }
 
 function updateSelectedCoursesBar() {
--- a/src/js/hyperschedule.js
+++ b/src/js/hyperschedule.js
@@ -29,8 +29,8 @@ const courseSearchScheduleColumn = document.getElementById("course-search-schedu
 const courseSearchInput = document.getElementById("course-search-course-name-input");
 const courseSearchResultsList = document.getElementById("course-search-results-list");
 
+const selectedCoursesColumn = document.getElementById("selected-courses-column");
 const importExportDataButton = document.getElementById("import-export-data-button");
-
 const printButton = document.getElementById("print-button");
 
 const courseDescriptionBox = document.getElementById("course-description-box");
@@ -640,7 +640,8 @@ async function retrieveAPI(endpoint)
 
 function attachListeners()
 {
-  document.addEventListener("DOMContentLoaded", updateCourseSearchBar());
+  document.addEventListener("DOMContentLoaded", updateCourseSearchBar);
+  document.addEventListener("DOMContentLoaded", updateSelectedCoursesBar);
 
   courseSearchToggle.addEventListener("click", displayCourseSearchColumn);
   scheduleToggle.addEventListener("click", displayScheduleColumn);
@@ -664,6 +665,7 @@ function attachListeners()
   });
   window.addEventListener("resize", updateCourseDescriptionBoxHeight);
   window.addEventListener("resize", updateCourseSearchBar);
+  window.addEventListener("resize", updateSelectedCoursesBar);
 
   // Attach import/export copy button
   let clipboard = new Clipboard("#import-export-copy-button");
@@ -1141,7 +1143,9 @@ function updateCourseSearchBar() {
   const helpButtonWrapper = document.getElementById("help-button-wrapper");
   const helpButton = document.getElementById("help-button");
 
-  let tableValue = "table-cell"
+  // default value
+  let tableValue = "table-cell";
+
   if (courseSearchColumn.offsetWidth < 
     (courseClosedToggleLabel.offsetWidth + helpButton.offsetWidth + 100)) {
     tableValue = "table-row";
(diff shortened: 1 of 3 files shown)


src/js/hyperschedule.js
```diff
@@ -48,12 +53,14 @@ const importExportSaveChangesButton = document.getElementById("import-export-sav
 const importExportCopyButton = document.getElementById("import-export-copy-button");
 
 //// Global state
+const gGreyConflictCoursesOptions = ["none", "starred", "all"];
```
raxod502: This isn't global state if it's a constant. I would suggest putting it with the other data constants.

src/js/hyperschedule.js
```diff
@@ -440,10 +454,37 @@ function generateCourseDescription(course)
...
+        if (course.starred && courseConflictWithSchedule(course)) {
+          hue = "monochrome";
+          seed = "-10";
+        }
+        break;
+
+      case gGreyConflictCoursesOptions[2]:
+        if (courseConflictWithSchedule(course)) {
+          hue = "monochrome";
+          seed = "-10";
```
raxod502, line 471: Where did `-10` come from? Shouldn't we just hardcode a specific color rather than randomly generating always the same color?

src/js/hyperschedule.js
```diff
@@ -655,13 +720,34 @@ function attachListeners()
...
   });
   selectedCoursesList.addEventListener("sortstop", () => {
     gCurrentlySorting = false;
   });
+
+  // for (var i = 0; i < conflictCoursesRadios; i++) {
+  //   conflictCoursesRadios[i].addEventListener("click", () => {
+  //     gGreyConflictCourses = gGreyConflictCoursesOptions[i];
+  //   })
+  // }
```
raxod502: What was wrong with this? Did you just forget `.length`?

src/js/hyperschedule.js
```diff
@@ -1172,90 +1279,37 @@ function updateSelectedCoursesBar() {
...
-function updateSelectedCoursesBar() {
-  const githubLink = document.getElementById("github-link");
-  const importExportButtonWrapper = document.getElementById("import-export-data-button-wrapper");
-  const printButtonWrapper = document.getElementById("print-button-wrapper");
-
-  // default values
-  let tableValue = "table-cell";
-  let floatValue = "right";
-  let marginValue = "0 auto";
-  let printPaddingLeftValue = "10px";
```
raxod502, line 1241: Why did this code get deleted?

raxod502 (changes requested):
Thanks, this is a really great feature that people have been requesting for a long time. I have a few questions about the code but after they are addressed this looks good.

src/js/hyperschedule.js
ingridt123: I'm not sure -- I tried the following and it didn't seem to work. I think it's to do with i=3 when the event is triggered.

`for (var i = 0; i < conflictCoursesRadios.length; i++) {
    conflictCoursesRadios[i].addEventListener("click", () => {
      gGreyConflictCourses = greyConflictCoursesOptions[i];
      toggleConflictCourses();
    })
  }`

src/js/hyperschedule.js
ingridt123: I just did this so that different color formats could be returned and I'm unsure how else to implement this?

src/js/hyperschedule.js
ingridt123: I think somewhere along the way I messed up the merges so this code was duplicated. The most updated code is still in the hyperschedule.js file.

src/js/hyperschedule.js
raxod502: Looks like [an obscure misfeature of JavaScript](https://codeburst.io/asynchronous-code-inside-an-array-loop-c5d704006c99). Try using `let` instead of `var`. In fact, `let` is basically always preferred to `var`, for this and other reasons.

src/js/hyperschedule.js
raxod502: Okay, I'll trust that what you have in this pull request is what we should have in master.

src/js/hyperschedule.js
raxod502: Oh, I see. I didn't realize `format` could be different. Fair enough :)

raxod502:
The settings button looks a bit inconsistent with the other buttons on Firefox in KDE Linux:

image

Do you think you could fix that?

src/js/hyperschedule.js
ingridt123: This should be resolved in the newest commit

ingridt123:
> The settings button looks a bit inconsistent with the other buttons on Firefox in KDE Linux:
> 
> image
> 
> Do you think you could fix that?

This should be fixed in the newest commit
Screen Shot 2019-08-13 at 9 20 46 AM

raxod502:
Okay, mostly looks great. However, in testing, I'm unable to see anything grayed out. As you can see, I've added a conflicting course but it's rendered in the same color as usual:

image

Any idea what could be wrong?

ingridt123:
The newest commit should fix that.
Also I added a `updateCourseDisplays()` function since I found that many of the functions were calling the same three functions to update search results, selected course list and schedule.

raxod502:
It works now for graying out all conflicts, but not for graying out conflicts with starred courses:

image

ingridt123:
There doesn't seem to be any conflicts in the courses selected above?
I tried it by adding the Theoretical Mechanics course, which conflicts with Intro to Africana Studies and it seems to be working. In the first screenshot, TheoMec is not checked, and it is checked in the second screenshot (and the same greying out is shown for using the "grey out all conflicting courses" option).
Screen Shot 2019-08-15 at 9 49 54 AM
Screen Shot 2019-08-15 at 9 49 42 AM

AmitHarlev:
This isn't very important, but I think it would look better without the "Gray out conflicting courses" text in the settings menu.

ingridt123:
> This isn't very important, but I think it would look better without the "Gray out conflicting courses" text in the settings menu.

I am open to going either way, but I just put this heading so it will be clear when there are more settings added in the future.

raxod502:
@ingridt123 If you have "gray out conflicts with starred courses" selected, then shouldn't courses in the course search results be grayed out if they conflict with selected courses that are starred?

ingridt123:
> @ingridt123 If you have "gray out conflicts with starred courses" selected, then shouldn't courses in the course search results be grayed out if they conflict with selected courses that are starred?

By the "gray out conflicting starred courses" option, I meant for it to gray out courses that were starred and conflicted with other courses, so courses that are not starred would not be grayed out even if they are conflicting with other courses.

raxod502:
Isn't it strictly less useful for only starred courses to be grayed out? I assumed that the primary use case was if you starred courses, then those are the ones you've registered for, and you'd then want to search for additional courses to add that don't conflict with the ones already on your schedule.

ingridt123:
Ok I think that makes more sense.
Just to clarify, perhaps instead of "only gray out conflicting starred courses" to have "only gray out courses that conflict with starred courses"?

src/js/hyperschedule.js
```diff
@@ -54,6 +59,7 @@ let gApiData = null;
 let gSelectedCourses = [];
 let gScheduleTabSelected = false;
 let gShowClosedCourses = true;
+let gGreyConflictCourses = greyConflictCoursesOptions[1];
```
raxod502: I don't think this can default to being enabled, because the feature you're adding introduces a considerable performance regression.

src/js/hyperschedule.js
```diff
@@ -1690,6 +1744,9 @@ function readStateFromLocalStorage()
   gShowClosedCourses = readFromLocalStorage(
     "showClosedCourses", _.isBoolean, true
   );
+  gGreyConflictCourses = readFromLocalStorage(
+    "greyConflictCourses", _.isString, greyConflictCoursesOptions[1]
+  );
```
raxod502, line 1748: I feel like this should be validated.

raxod502 (changes requested):
Yes, that sounds fine. I'll wait for that change.

src/js/hyperschedule.js
ingridt123: Sorry what do you mean by this?

raxod502: If there's an invalid value in the local storage, we shouldn't just assign it to `gGreyConflictCourses`. Otherwise it becomes much harder to reason about the code. If the value is invalid we should just reset to the default.

src/js/hyperschedule.js
ingridt123: This should be resolved in the newest commit, where the default is no graying out conflicting courses.
`let gGreyConflictCourses = greyConflictCoursesOptions[0];`

src/js/hyperschedule.js
ingridt123: I think this is resolved in the newest commit:
`gGreyConflictCourses = readFromLocalStorage(
    "greyConflictCourses", _.isString, greyConflictCoursesOptions[0]
  );`

ingridt123:
Updated the graying out functionality in the latest commit, to "Only gray out courses conflicting with starred courses" and "Gray out courses conflicting with any course".

src/js/hyperschedule.js
raxod502: I am pretty sure that with this code, if `greyConflictCourses` has a value of `"asdf"` in local storage, then that value will just get assigned to the global variable.

ingridt123: Ahh sorry I didn't quite understand before.
This should be resolved in the newest commit.

src/js/hyperschedule.js
```diff
@@ -343,6 +349,13 @@ function courseToInstructorLastnames(course)
   return course.courseInstructors.map(fullName => fullName.split(",")[0]).join(",");
 }
 
+function coursesEqual(course1, course2) {
+  if (course1.courseCode == course2.courseCode) {
+    return true;
+  }
+  return false;
```
raxod502: [We should never use `==`.](https://2ality.com/2011/12/strict-equality-exemptions.html) I would also suggest just refactoring this whole function to `return course1.courseCode === course2.courseCode`.

raxod502 (changes requested):
Looks great! One final comment before I can merge.

src/js/hyperschedule.js
ingridt123: This should be resolved in the newest commit.

raxod502 (approved):
Alright, looks good. Thank you for your patience!

PR #90, opened 2019-08-13
Revert "Updated to newest version of Ionicons"

ingridt123:
Reverts MuddCreates/hyperschedule#88
The new ion icons are slightly different so this is probably a better fix (was going to switch to updated version for some icons that aren't in old version).


raxod502 (approved):
:leftwards_arrow_with_hook:

PR #91, opened 2019-08-15
Changed how courses are added to schedule...

ingridt123:
...so that when multiple conflicting starred courses are added, the course highest on the selected courses list is "on top" (shown in schedule)

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -96,7 +96,7 @@ body {
 }
 
 #schedule-column {
-    height: 93%;
+    /*height: 93%;*/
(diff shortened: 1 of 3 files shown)


raxod502:
Since this branch is based on the one from your other pull request, I can't review the changes independently. If want to fix it, you may want to read about `git rebase`, which you can use to take the commits from this branch and move them so that they branch directly off of `master` instead of off of your first branch. Otherwise, you can wait until we figure out your first pull request, since once it's merged the problem will go away. If the latter, just ping me if I forget about this pull request.

PR #93, opened 2019-08-16
Added Folders Feature

AmitHarlev:
This is a feature I wanted last semester when trying to make my schedule. This adds collapsible folders where only one course is added to the schedule from each folder (staring courses can override this) following the same logic as the selected course list. I didn't know where to place the add folder button without it being awkward, so I added it to a menu which is located by right clicking the selected course list.

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -360,6 +360,32 @@ body {
     display: none;
 }
 
+/* Folder Listing */
+
+.folder-box {
+    cursor: pointer;
+    margin: 0 auto;
+    padding: 6px 0;
+    position: relative;
+    width: 95%;
+    font-size: 0.8em;
+}
+
+.folder-box-content {
+    display: block;
+    margin: 0 auto;
+    position: relative;
+    padding-left: 5px;
+}
+
+.folder-box-button {
+    background-color: transparent;
+    border: none;
+    cursor: pointer;
+    display: none;
+}
+
+
 /*** Course search listing ***/
 
 #course-search-column .course-box-select-label {
@@ -410,7 +436,7 @@ body {
     text-align: center;
 }
 
-#selected-courses-column .course-box-collapse-label {
+#selected-courses-column .folder-box-collapse-label {
     display: block;
     height: 50px;
     margin-top: -25px;
@@ -427,7 +453,7 @@ body {
     font-size: 1.2em;
 }
 
-#selected-courses-column .course-box-collapse-icon {
+#selected-courses-column .folder-box-collapse-icon {
     font-size: 1.2em;
 }
 
@@ -476,6 +502,17 @@ body {
 }
 
 #selected-courses-column .course-box-remove-button {
+    display: block;
+    height: 50px;
+    margin-top: -25px;
+    line-height: 50px;
+    position: absolute;
+    right: 0;
+    top: 50%;
+    width: 25px;
+}
+
+#selected-courses-column .folder-box-remove-button {
     display: block;
     height: 50px;
     margin-top: -25px;
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -222,7 +222,7 @@ body {
 
 #course-description-box p {
     font-size: small;
-    lpine-height: 100%;
+    line-height: 100%;
     margin: 0px 10px;
 }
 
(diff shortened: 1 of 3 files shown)


AmitHarlev:
I am leaving this as a draft for now for two reasons. One is that I have not yet had a chance to review my code and do even minimal cleanup. The other is that this feature will break hyperschedule if merged with my previous pull request. I will wait until that PR is accepted and then add a small commit that fixes that issue before opening this PR for review. For now I just wanted you to be able to see that this existed and let me know any changes you may want.

AmitHarlev:
Also it seems like everything is slowed down a lot because how it is written right now. I hope that it will get a lot better when I clean up my code.

AmitHarlev:
The merge request broke things so this isn't ready yet.

AmitHarlev:
It now works and is ready to be reviewed.

raxod502:
I think a bit of UX work is needed for this. I like the feature, but when I tried to use it, it took me several minutes to figure out how.

- would it be possible to make it so you can drag and drop into a folder directly? I tried that a bunch before eventually realizing that you have to click the folder icon.
  - if so, then perhaps "new folder" can just be a button at the bottom of the selected courses list
- this feature, and how it interacts with scheduling, should be documented in the help modal.
- right-clicking just isn't discoverable enough. Why not add an option to the dropdown menu you get when you click on the folder icon? It's pretty obvious that if you want to do stuff with folders, you should click on the folder.
- can we surface a bit of text that makes it clear what is going on when you click on the folder menu?
- I think it would be much nicer if you're prompted for a name when you create a folder.

I might have more feelings after I look at the new version. But I feel some iteration is needed.

kwshi:
I'm also not a huge fan of the current visual appearances.
image

A few aesthetic suggestions:
- The font sizes are kind of all over the place, as are the font styles (some bold, some not?).  I think the presence of the dropdown arrow is enough to indicate that something is a folder.  Having different font sizes looks messy.  I suggest keeping everything there the same font size, and color. 
  - Regarding font size for the nested course item: I believe the best way to do this is to replace the `0.8em` font size in `.course-box` with `0.8rem`
  - Regarding font size/color for the folder: I think you just need to add an explicit `font-size: inherit` on the folder item, since HTML inputs don't do that by default.
- The spacing between the dropdown arrow and the folder name is too much.
- That bullet point on items inside a folder looks a little messy; I think the indentation is sufficient to convey that the course is in the folder.  Regarding the folder dropdown triangle icon: I think it would look nicer without the black circular background; the black circle looks a little "noisy" from a design perspective, in my opinion.
- The spacing is inconsistent between folder-item vs other items.  Again, I think the indentation is sufficient to convey that an course is in a folder, and I think it would look cleaner and more consistent if the vertical spacing between items were uniform throughout.
- I also think it looks messy that the crosses on the right are not aligned.  I think the left indent is sufficient, and it would look cleaner to keep the right edges aligned.
- Let's make the folder icon color black, consistent with the rest of the icons, shall we?

Here's a mock-up of what I think might look good (I also added a light gray bar to the folder, because, I don't know, I felt like it?):
image

Meanwhile, a few UX suggestions:
- When creating a new folder, having a whole modal pop up to prompt for a folder name feels a little "loud".  I think it might be nicer if, instead, it just focused on the (editable) folder name input (with a white background to make it obvious that it's editable).  Something vaguely like this:
  image
  In fact, it would be nice if the background were white or something to indicate it's being edited _whenever_ it is focused.
- I'm also a fan of using drag-and-drop, instead of a folder icon with a dropdown menu, for assigning courses to folders.  I think it's much more intuitive to have this behave like Photoshop's layers menu, with one global "New folder" button, and drag-and-drop to assign courses to different places.
- On naming: I think it's more straightforward to call these things "groups" instead of "folders"--folders sounds more like something I do with files on my computer rather than courses on Hyperschedule.  The most intuitive name seems to be "group" to me, because essentially it is just a "group" of courses that are ordered together, etc.

Also, I noticed that the folder element has `id="folder"`.  That should almost certainly be a `class="folder"` instead (it's incorrect to have multiple elements with the same `id`s).

kwshi:
Also, there have been bug fixes/new changes made to `master`, so it's probably a good idea to merge those changes into this PR first before accepting this PR.

raxod502:
This thread is being closed automatically by [Tidier](https://github.com/raxod502/tidier) because it is labeled with "waiting on response" and has not seen any activity for 90 days. But don't worry—if you have any information that might advance the discussion, leave a comment and the thread may be reopened :)

PR #98, opened 2019-08-19
Changed how courses are added to schedule...

ingridt123:
...so that when multiple conflicting starred courses are added, the course highest on the selected courses list is "on top" (shown in schedule)

I think I fixed the problem with git rebase as well


raxod502 (approved):
:confetti_ball:

PR #100, opened 2019-08-22
 Fixes problem where course search bar uses table-row instead of table-cell if reload on schedule page

ingridt123:
This happens because if DOMContentLoaded is used, offsetHeight of elements in course search bar are all 0.


raxod502 (approved):
👓

raxod502:
Just a tip, for commit messages you may want to look at https://chris.beams.io/posts/git-commit/ which is considered more or less the standard :)

PR #101, opened 2019-08-23
Adds option to print only starred courses

### LINKED ISSUE #76: No way to print only starred courses
Currently when you print your schedule, all the courses are printed, even the non-starred ones. There should be a way to print only the starred courses.

ingridt123:
Adds a dropdown so that the user can choose to print all courses or only starred courses

Resolves: #76


raxod502 (approved):
:star2:

PR #102, opened 2019-08-23
Add search argument to filter by department

### LINKED ISSUE #65: Search by department
It is currently a little annoying to search for PE courses, since the string "PE" matches many courses that are not PE courses. It would be nicer if there were a specific way to request courses in a particular department.

AmitHarlev:
Filter by department by adding "dept:" as an argument in your search.
Example: dept:math analysis

Resolves #65


raxod502 (approved):
:mount_fuji:

PR #103, opened 2019-08-24
Clarifies modal window header for graying conflicting courses

ingridt123:
Changed to "Gray out courses conflicting with schedule"


raxod502 (approved):
:scroll:

PR #106, opened 2019-08-24
Updates search filter function and adds instructions to "How It Works" modal

ingridt123:
- Refactors the search filter code and adds filter for searching by college
- Lists the search filter keywords in the "How It Works" modal


AmitHarlev:
There is a bug. I tired "dept:math" and got nothing back. Just "dept:" by itself returns everything but adding anything past that returns nothing.

AmitHarlev:
The problem seems to be capitalization. If you do "dept:MATH" it works. It should work with any capitalization. Also right now in the production hyperschedule "Dept:math" doesn't work for similar reasons. It would be nice if you could have this PR fix both of those issues and make capitalization not affect anything.

ingridt123:
The capitalization problems should be resolved in the latest commit.

kwshi:
A few suggestions:
- I think calling the filter-by-college _filter-by-school_ and using a prefix `sc:` or `sch:` instead sounds nicer to me.  That's just a matter of preference, so feel free to ignore this suggestion if you do feel that `col:` is nicer.
- The "how-to-use" note is nice: I think it'll look nicer if you organized the options into a list, something like
  > You can specify additional filters in the form of `key:text` to filter on more specific parameters.  The following filters are supported:
  > - `dept`: filter by department (e.g. `MATH`, `CSCI`, etc.; case insensitive)
  > - `sch`: filter by school (e.g. `HM`, `SC`, etc.; case insensitive)

kwshi:
Also, new changes/bugfixes have been made to `master`, so maybe merge in those changes before we include this PR?

ingridt123:
These changes have been resolved in the latest commit.

raxod502 (changes requested):
* Can we make `s`, `sc`, `sch`, `school`, `college`, `c`, etc. into aliases of `col`? I feel like it's hard to remember exactly what abbreviation of what word is used. We don't have document this but it's nice to have.
* I think we've got some room in the placeholder search text for an extra tip like `(or filter with dept:csci, school:hm)`. What do you think?

kwshi:
> * Can we make `s`, `sc`, `sch`, `school`, `college`, `c`, etc. into aliases of `col`? I feel like it's hard to remember exactly what abbreviation of what word is used. We don't have document this but it's nice to have.

I'm not opposed to it, though I also think users can also just pull up the "how-to" again if they forget (although maybe that might be annoying?).  I'm not exactly sure about the one-letter `s` and `c` aliases though--I think those seem a little too short, and if we introduce new filters in the future, it might seem weird that these have one-letter aliases but others don't?

As a matter of fact, I have suggestion to throw out (see what y'all think): we document the full key name `school`/`college` (whichever one works; shouldn't be too hard to remember), and also implicitly allow the other less-obvious aliases such as `sc`/`col`, etc.

>  * I think we've got some room in the placeholder search text for an extra tip like `(or filter with dept:csci, school:hm)`. What do you think?

I like this!

raxod502:
Yes, documenting the longer versions is ideal, I think.

> weird that these have one-letter aliases but others don't?

Fair enough.

> users can also just pull up the "how-to" again

You forget that the overwhelming majority of users are not even aware that this screen exists, a fact of which I was made painfully aware from feedback :)

kwshi:
Also, I believe it was discussed in the past that implementing GitHub-style search strings, i.e. the feature implemented by this PR, could be used to filter on criteria like closed vs. open courses.  Two related questions, then:
- Should we implement that (seems nice, why not, you ask?  see below)?
- If we do, there would be some feature overlap with the now-existing "Filter -> Show closed courses" checkbox--a checkbox which, at one point, some people expressed interest in _replacing_ with GitHub-style search strings, I believe.  In that case, what to do?  Should we keep both?  Should we get rid of the Filter menu and rely only on the search strings?  Should we avoid the search strings for closed/open?  Should we have both?

(And the same questions may be asked of the other two Filter menu options for hiding/showing conflicting courses.)

kwshi:
Other things we could filter on:

- Instructors (currently, the built-in string filter matches course name OR description OR instructor); we could change that behavior to only match on name/description (which seems somewhat more intuitive to me) and have an `instructor:`/`instr:` key to search by instructor name.
- Credits: might even be interesting to consider supporting "ranges", e.g. `credits:>1`?
- Semester: first-half/second-half/both/whatever?
- Meeting days: support some syntax to include/exclude certain days?
- Meeting time: support some range-based syntax to include/exclude certain times?

(Basically everything on the API, lol).  Anyway, some of these are complex-ish and don't have to be implemented all in this PR, but they're worth thinking about to see what we want to have, what we don't want, etc.

ingridt123:
> Also, I believe it was discussed in the past that implementing GitHub-style search strings, i.e. the feature implemented by this PR, could be used to filter on criteria like closed vs. open courses. Two related questions, then:
> 
> * Should we implement that (seems nice, why not, you ask?  see below)?
> * If we do, there would be some feature overlap with the now-existing "Filter -> Show closed courses" checkbox--a checkbox which, at one point, some people expressed interest in _replacing_ with GitHub-style search strings, I believe.  In that case, what to do?  Should we keep both?  Should we get rid of the Filter menu and rely only on the search strings?  Should we avoid the search strings for closed/open?  Should we have both?
> 
> (And the same questions may be asked of the other two Filter menu options for hiding/showing conflicting courses.)

I think having the filter menu options are useful so that users can keep the same settings on without having to retype it in the search every time they use the app.
But I agree implementing GitHub-style search strings would also be useful. I guess this could be implemented with something like like `closed:yes` and `closed:no`?

ingridt123:
I have added to the placeholder text in the search bar and implemented the aliases for the text filters.

Also, I will work on the other filters and implement them in later PRs.

raxod502:
A potential way to support both use cases is to have the "filters" button pop up a modal dialog which has quick checkboxes for common operations like "hide conflicting courses", "hide closed courses", etc. Manipulating these checkboxes would automatically populate a text input field in the modal, which uses GH-style filters. The idea is then that this text input is automatically concatenated with the search query, so you can use it to enable filters persistently. The advantage of this approach is that you don't have to worry about the GH-style filtering if you don't understand it, because you can configure things through the interface, but you are incrementally exposed to the advanced search syntax.

AmitHarlev:
If this PR is currently functional (no bugs), would it be possible to merge it. I would like to contribute to these filters since I was originally planning on making filters based on time after making the dept one but was sidetracked by the changes requested for the folders. If this one is merged, feature functionality can easily be added in smaller PR's by multiple people instead of everything being done in a single PR.

kwshi:
I think it looks good to me!  (Sorry I didn't get this sooner; end-of-summer/travel has been a little hectic.)  Good work!

PR #110, opened 2019-08-25
Implement virtualized scrolling

### LINKED ISSUE #104: Many duplicate courses; related to performance issues

kwshi:
Re-resolves #104.

Currently, the course search results list is a "classic" infinite/incremental scroller, wherein about two screen-heights' worth of DOM is rendered at first (as opposed to all 2000+, one for each course), and more is rendered as the user scrolls down the results list.  There's a few downsides to this approach:
1. The scrollbar acts janky as new items are rendered--every time you scroll down, new items are added, and the scroll area becomes taller, so the scroll bar "teleports" up.  Long-time users of infinite-scrollers are accustomed to this jankiness, but it's not a good thing to be accustomed to.  Related:
  - The scrollbar provides no estimate as to how large the total list should be, or how far into the list the user currently has scrolled (not that this is critical information to have, but it's a nice UI touch that is missing with this approach).
  - There's no nice way to "skip" to the bottom of the scroll list: you have to scroll down, let new items render and the scrollbar teleport, then scroll again, and again, until eventually you reach bottom.  Scrolling experience: 2/5.

  hyperschedule-janky-scroll
2. Performance gets worse the farther you scroll down, since new elements are rendered in but no elements are removed, so the DOM becomes more and more bloated (despite the same, fixed amount of content being actually visible on screen).  By the time you scroll to the bottom of the 2000+ course list, latency is back up to 1+ seconds.

  hyperschedule-scroll-deterioration

3. (Not an issue with infinite scrolling, but a bug the current behavior: previously, the course list was completely broken, loading in multiple duplicate copies of the same courses; see #104.  I _thought_ I had fixed that in #105, but it turns out I must not have looked at the incremental rendering logic carefully enough, because the performance issues were fixed, but the duplicate courses are very much still there!)
4. (Another issue with the current implementation: filtering courses based on the search input and re-rendering based on scrolling is happening in the same step, `updateCourseSearchResults`.  That's inefficient, since it means that the courses are getting re-filtered unnecessarily every time the course list is scrolled; they should be re-filtered only when the search input changes.)

As the name implies, _infinite_ scrollers are meant for _infinite_ data (or at least data where you don't readily know how much total stuff there is); in those situations, it makes sense that you can't estimate the total length or skip to the end, because you _don't know_ the total length or the end!  The good news is, we have _finite_ data, so we can use a different rendering technique, termed by some a "virtualized list".  Here's how it works:
- Render a single dummy list item on DOM load, and measure its height.
- Compute the total height of all list items (number of items * height of individual items).  Insert a blank background overlay/placeholder (`courseSearchResultsPlaceholder`) into the scroll container (`courseSearchResults`) so that it has the correct scroll height.
- On each scroll event (and the initial render), calculate, based on the scroll position, window viewport height, and the height of individual items, the indices of items that would be visible on the screen (and some additional few above and below, for padding).  Calculate their positions, and render _only those items_ to the screen, with the correct absolute positions set.
- The resulting effect is that the scroller _simulates the entire list_ being rendered, because it has the correct scroll height, while in fact _improving on_ the performance of lazy scroll rendering, because it always renders to the DOM only a small, fixed-size subset (~60 or so) of the list items instead of increasing the subset with the scroll position (causing performance to deteriorate as you scroll down).  Wow; the best of both worlds!

A key observation as to why this technique works is that, currently, the performance bottleneck largely lies in having a cluttered DOM (and not network loading speeds, or in-memory data stores).  From my experience, I've found that the app starts to feel laggy as soon as you have ~1k course entity items _rendered to DOM_.  When fewer elements are rendered to DOM, even if we had ~10k course objects in our API data, the performance would still be fairly snappy.  Thus, by reducing the DOM size, this technique dramatically improves UI performance.

Here's how this PR fixes each one of the above mentioned limitations with the current approach:
1. Janky scrollbars are not an issue anymore, because the scroll height is calculated correctly once and does not vary throughout scrolling, so the scrollbar is an accurate, meaningful reflection of the actual scroll heights and positions.

  hyperschedule-scroll-virtual-nice

2. The number of elements rendered varies not with scroll position but the height of the viewport, so the same (small) number of elements is rendered regardless of where you scroll.  Hence scrolling down causes no performance deterioration.

  hyperschedule-scroll-virtual-no-deterioration

3. The entire list rendering logic has been redone.  It is now actually much simpler to reason about and hence less bug-prone.  As a side effect, the duplicate courses bug is no more.
4. This implementation splits up re-rendering (on window resize, list scroll events) and re-filtering (on search change), and only re-filters when necessary.

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -34,11 +34,9 @@
 }
 
 .toolbar {
-    display: table;
-    margin: 0 auto;
+    display: flex;
     padding: 0;
-    table-layout: fixed;
-    width: 95%;
+    margin: 0 1em;
 }
 
 /** Page layout **/
@@ -72,16 +70,13 @@ body {
 
 #course-search-schedule-column {
     padding-top: 20px;
+    display: flex;
+    flex-direction: column;
 }
 
-#course-search-toggle-wrapper {
-    display: table-cell;
-    vertical-align: middle;
-}
-
-#schedule-toggle-wrapper {
-    display: table-cell;
-    vertical-align: middle;
+#course-search-toggle-wrapper, #schedule-toggle-wrapper {
+    flex-grow: 1;
+    flex-basis: 0;
 }
 
 #course-search-toggle {
@@ -107,11 +102,29 @@ body {
 /** Course search **/
 
 #course-search-column {
-    height: 100%;
+    flex-basis: 0;
+    flex-grow: 1;
+    display: flex;
+    flex-direction: column;
+}
+
+#course-search-column.hidden {
+    display: none;
+}
+
+#course-search-results {
+    overflow-y: auto;
+    flex-basis: 0;
+    flex-grow: 1;
+    position: relative;
 }
 
 #course-search-results-list {
-    overflow-y: auto;
+    height: 0;
+}
+
+#course-search-results-placeholder {
+    position: relative;
 }
 
 /*** Course search header bar ***/
@@ -176,11 +189,12 @@ body {
     background-color: #b8b8ff;
 }
 
-/*** Course search placeholder ***/
+/*** Course search end of list ***/
 
-#course-search-placeholder {
-    margin: 15px;
+#course-search-results-end {
+    margin: .5em 1em 1em 1em;
     text-align: center;
+    color: gray;
 }
 
 /** Selected courses **/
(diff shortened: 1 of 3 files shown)


raxod502 (approved):
image

PR #121, opened 2019-11-19
Added Buttons To Minimize/Close Course Description

nguyenank:
I got sick of really long course descriptions making it difficult/impossible to scroll through my selected classes, so I created buttons to minimize or close the description. The buttons are hidden by default, only appearing when a course is clicked on, and disappear again if the close button is clicked. I tried to make the two buttons side-by-side the same size as the selected courses below the description.

I don't know CSS very well so the formatting is a little bit hacked-together (I manually set the background color to match the rest of the primary buttons, which is a little sketchy). I also definitely learned JS doing this so my code probably isn't the cleanest it could be.

Also, completely unrelated to the buttons, I switched the filter dropdown to be right-aligned because I never knew what they did because the text was always partially hidden by the right column; so now I know what they do, at least.

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -253,9 +253,22 @@ body {
 
 /*** Course description box ***/
 
+#minimize-close-wrapper {
+  display: flex;
+  margin: 0 1rem 0 1rem;
+}
+
+.minimize-close {
+  display: none;
+  background-color: #307cf7 !important;
+  font-size: small;
+  flex-grow: 1;
+}
+
 #selected-courses-wrapper {
   overflow-y: auto;
 }
+
 
 #course-description-box-outer {
   display: block;
@@ -565,9 +578,11 @@ body {
 #selected-courses-column .course-box-remove-button {
 }
 
+/*
 #selected-courses-list {
   margin-top: -0.5rem;
 }
+*/
 
 /** Modals **/
 .modal-input-text {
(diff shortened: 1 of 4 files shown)


raxod502:
That's very handy, I like it! Don't forget to run Prettier; as you can see the CircleCI build is failing.

kwshi:
I like this!  

I feel the two buttons (hide and close) are a bit redundant, though--both achieve the (approximately same) effect of minimizing the course description box, and clicking on a course re-opens the course description box in both cases.  

The only difference between the two buttons is "hide" allows you to reopen the box, and "close" doesn't.  Why not do away with one of these buttons / make them more different?

nguyenank:
That's a fair point; I think I'll probably do away with the close button because its a bit less useful. I suppose the alternative to differentiate them would be to make it so if minimized, the course description changes on clicking a course but doesn't open, but I can't really imagine too many scenarios where you wouldn't want the course description when clicking on a course and switching course descriptions behind the scenes seems unintuitive.

kwshi (changes requested):
Looks nicer.  I started looking at the code a bit and was wondering whether you could make things simpler--instead of hiding the description box by `removeChild`ing all course description elements, `appendChild`ing to a `gMinimized...` container, then doing the reverse when unhiding, why not just set the course description box height to 0px?

nguyenank:
That's true and incredibly valid (and now changed, along with some additional minor code streamlining).

kwshi (dismissed):

kwshi:
Thanks for this!

Issue #107, opened 2019-08-24
Suggestion: use RE:DOM instead of document.createElement


kwshi:
Currently, we use hundreds of manual calls to `document.createElement`, `el.setAttribute`, `el.classList.add`, etc. to generate HTML elements and set properties.

[RE:DOM](https://redom.js.org/documentation) is an ultra-lightweight, efficient library for creating DOMs.  I think we can produce much nicer-looking code, and in fact more efficient code, using RE:DOM to create our elements instead of manual DOM calls.

As a quick demonstration, here's some existing code used to create a course checkbox:

```javascript
const selectToggle = document.createElement("input");
  selectToggle.setAttribute("type", "checkbox");
  selectToggle.classList.add("course-box-button");
  selectToggle.classList.add("course-box-toggle");
  selectToggle.classList.add("course-box-select-toggle");
  selectToggle.checked = !!course.selected;
```

and here's how it would look using RE:DOM:

```javascript
el('input', {
  type: 'checkbox',
  className: 'course-box-button course-box-toggle course-box-select-toggle',
  checked: !!course.selected
});
```

A few nice things about using RE:DOM:
- It operates on plain, native DOM types, so it's _really_ easy to try porting parts of the app to RE:DOM incrementally, and we don't have to do a massive rewrite of all the UI code.
- RE:DOM supports a declarative list renderer that allows you to pass in a list of data objects (i.e. courses), and it does the diffing and efficient updating for you (much like React, but more lightweight and compatible with vanilla JS), which is almost certainly more efficient than what we do now, which is deleting all elements inside courseResultsSearchList and re-appendChild'ing every single time a scroll event happens, etc.
- The code is much more readable.

raxod502:
Sounds like a great idea to me. I'd also be okay with using JSX (I believe we could even enable this in the Babel configuration right now without having to change anything else), which I think is also pretty readable and easy to understand for the unfamiliar. Either option is good.

kwshi:
Long since obsolete, as the new frontend (V2) is getting rewritten in React.

PR #124, opened 2020-01-02
Groups Feature

nguyenank:
This is heavily heavily based on the folders feature Amit pioneered in a different pull request, and the subsequent conversations thereafter about UX stuff, especially Kye's mockup.

The big change in relation to original Hyperschedule is that I changed the JS library used for the drag-and-drop in the selected courses list.

For scheduling purposes (and all other purposes that use selected courses like printing and stuff), it basically acts as the groups didn't exist and the selected courses list was flattened. 

One quirky thing is that naming/renaming is cancelled whenever focus is away from the group name. That includes when course data is retrieved. So you get automatically booted from typing almost immediately to 5 seconds after you start, depending on the timing. However, I didn't really like any alternative cancellations besides losing focus, but also didn't know how to deal with the async function.


raxod502:
Deselection is a perfectly reasonable hook to use for canceling text input, but the five-second loss of focus is really, really bad. Can't we add some code to the function that updates the selected course list which maintains the user's selected text input? I can't imagine that the loss of focus is technically necessary.

raxod502:
That said, I will defer to @kwshi

raxod502:
Oh and also, I should really be able to finish my text input by pressing return. It's super confusing otherwise.

kwshi:
Sorry it took a while to respond!  As it stands right now, it doesn't seem to be working correctly.  A few comments, in approximate order of significance:
- The new drag-and-drop library doesn't seem to work at all on Firefox.  Perhaps someone else could try with Firefox and see if it's just me?  (Also could you point me to a website or something for this new drag-and-drop library?)
- Agree that the 5-second loss of focus renders it practically unusable.  It's even worse if you happen to focus just right before new course data gets loaded--then you get booted immediately.  That would have to be fixed before this can be merged.  (I suspect it happens on API reload because we blindly replace/recreate the course elements every time, instead of doing some smarter DOM-diffing thing--this could be improved using RE:DOM, see #107.)
- Possibly related to above/broken drag-and-drop, but the group renaming didn't work for me either.  I'd edit the text field, then hit enter or drop focus, but as soon as I lost focus the folder name was blank again.
- UX suggestion--I find the current UI slightly unintuitive--it wasn't clear to me at first what the text field for the group was.  In fact, at first I tried to search course names/type course codes into the field to add things to it (I'll admit part of the confusion was because drag-and-drop was broken).  I imagine adding a few "hint" elements to the design, like a "Group name..." placeholder to the text field or a default name of "Unnamed group", and maybe a small note inside empty groups saying "Drag courses to add..." (kind of like the "No results found" note for empty course search) would help make it easier to navigate how to use this feature.
- Another UX suggestion--I think it might be better to make the "Create group" narrower at the top right corner of the list and have it _not_ scroll with the courses, kind of like "new layer" and "new group" buttons in other applications with similar layouts, such as Photoshop/GIMP, etc.
- Appearance is slightly janky, at least on Firefox--when focused on the text field to rename a group, not-so-pretty 2000s-style borders appear.  That's an easy CSS fix, though.

nguyenank:
Alright, by the virtue of some time magically appearing in my schedule by reasons I don't want to think about, updates have been made. In order of previously presented concerns:

- I hacked together a fix for the loss-of-focus, which involves keeping track of what renaming boxes were open and the text that was selected, and checking for clicking anywhere but on the open text box (as well as enter/return). Based on some rudimentary testing, it felt decently smooth for me; the main issue was dragging to select text and going off the text box that would trigger ending; any ideas for that would be much appreciated.
- I have tested it on Firefox (after some changes were made, though), and drag-and-drop worked fine for me. I would definitely like someone else to test that, though. (main website for the new sortable library: https://github.com/SortableJS/Sortable)
- I did run into the renaming issue on Firefox, and it should be fixed now. (for the curious: it was related to Firefox still focusing on a disabled textbox, unlike Chrome)
- Groups now have a default name of Unnamed Group. I also added an "Drag courses to add" message to empty groups; however there is an issue with flickering as part of course refreshing I remove the messages from the DOM to prevent indexing errors (and they come back again shortly after). I'm also not completely sold on the message as currently when you drag in a new course it is around the message, which looks weird to me, and it seems vaguely cluttered to me with a lot of empty groups. Any opinions on this would be nice.
- Moved the Create Group button and made it look more like other buttons in the UX.
- Changed the text field appearance for Firefox.

Also, I don't know as much about the DOM-diff'ing thing, but given how annoying and unwieldy it is to add classes is to DOM elements, I would super be in favor of testing out RE:DOM.

nguyenank:
A few updates: I fixed the select text dragging issue, and scrapped the explanatory text entirely; I thought it was cluttered and the presence of the sideline for the groups (assuming dragging works) should be enough of a visual indicator. I upped the threshold to make dragging into a new group easier, too. 

I've been using this version to plan my schedule for next semester and noticed a few glitches (one case of course duplication, one of refusing to add a new course to a group) that I want to investigate (probably its related to the timing of resorting/regenerating the course list), and I have a few ideas for some later features (most notably a select group toggle to selected/deselected all courses in a group) but otherwise it works pretty much as I envisioned. The one thing is with import/export: it currently works, but basically invalidates all the groups. Stay tuned?

kwshi:
Just to get a sense of what this does: it makes it easier to drag and drop/reorder multiple courses at the same time--does it affect the actual way priorities/conflicts/schedules are computed?  Does putting something in a group cause it to get bumped if another course in the group is conflicting, etc.?  Or is it just to make dragging and dropping/organization easier?

nguyenank:
It does not affect priorities/conflicts/schedules; when computing those it essentially flattens the groups and then treats the list of courses as if the groups were not present. I did consider adding some sort of schedule interaction functionality, but I didn't really know what would be wanted in that regard, while I knew I at least personally wanted the organization/mass drag and drop functionality currently implemented.

Issue #130, opened 2020-02-22
Time/Day Search Keyword


JeremyTsaii:
Add-on to issue #20 “Search courses by time” submitted by PythonNut.

User Story: 

As a user, I want to filter classes by time of day and day of week so that I can find classes that fit my schedule in a more convenient fashion. Often times, I am left with a slot in my schedule that needs to be filled by a class with a specific start time and on a specific day. This feature would allow me to see courses starting at a specified time/day and easily let me fill such blocks in my schedule.

Implementation:

- [x] Add time keyword
- [x] Add day keyword

The user will be able to enter in a query like “11:00 AM” and/or “Monday” and classes that satisfy the start time/day will show up in the results. We also want the user to be able to input other keywords before/after the time/day keywords to further narrow their searches. 

![time keyword](https://github.com/JeremyTsaii/Images/blob/master/time.JPG)

![day keyword](https://github.com/JeremyTsaii/Images/blob/master/day.JPG)

This is part of CS121 Group 2. Team members include @JeremyTsaii, @godpng, @rory6103.

kwshi:
See #131 for some comments/suggestions on the query syntax.

PR #133, opened 2020-02-29
Implement "days:" Filter (Issue #130)

jacobyoo47:
Implements a new course search keyword filter "day:" that allows users to filter courses by day of the week:
newcoursesearch

The help modal was also changed to explain the new feature:
helpmodal

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -442,6 +442,10 @@
                 <span class="modal-input-text">col</span>: filter by college
                 (e.g. HM, SC)
               </li>
+              <li>
+                <span class="modal-input-text">days</span>: filter by class meeting days
+                (e.g. M, TR, MWF)
+              </li>
             </ul>
             <p>
               Click on the <b>+</b> button to add a course. Each section is
--- a/src/index.html
+++ b/src/index.html
@@ -444,7 +444,7 @@
               </li>
               <li>
                 <span class="modal-input-text">days</span>: filter by class
-                meeting days (e.g. M, TR, MWF)
+                meeting days (e.g. M, =TR, <=MWF, >MW)
               </li>
             </ul>
             <p>
--- a/src/index.html
+++ b/src/index.html
@@ -443,8 +443,8 @@
                 (e.g. HM, SC)
               </li>
               <li>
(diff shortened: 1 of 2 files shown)


kwshi:
Great work!  I think we can make it better.  Here are a few suggestions:
- Order shouldn't matter.  It looks like the filter simply checks whether the `days:` query string is a substring of the course's day string; that prevents queries like `days:fmw` or even `days:mf` from matching a `MWF` course.  It would be nice to make those behave correctly, _even if_ it's relatively unlikely that users will make queries like those.  That is to say, we should really be interpreting the days strings as _sets_ rather than _strings_, and check subsets instead of substrings.
- Now that we're talking sets, I'd also suggest allowing more sophisticated filters, like `<=MWF`, `>=M`, `=MW` to filter for "subset of", "superset of", or "exactly" (resp.).  A sample use case for `<=`: suppose I _don't_ want to go to class on Fridays; then I can search `<=MTWR` to search for classes that occur on a _subset_ of monday-thursday but not Friday.  As for `=` and `>=`, no examples immediately come to mind, but if we're going to implement `<=`, why not do the other two for consistency as well?  Finally, there should be a sane default for when no special operator is given, e.g. `days:MF`.  By the current behavior, the default seems to be `>=`.  I wonder if it should be `<=` or `=` instead?  (The rationale for `<=` is that it seems to be a more common/likely use case, per the above example; the rationale for `=` is that it's more of a "neutral" choice; if you have a good reason for `>=`, that's also good, as long as we are making this choice consciously.)  Up to you, but it might also be good to add these to the documentation as well.

jacobyoo47:
Thanks again for your suggestions!

We re-worked the days filter so that it uses sets instead of checking strings with the .match() method. Similar to the credits filter, we also implemented subset inequalities (as per your comment). If no inequality is given, the default behavior is to return schedules that contain a subset of the given days (>=). 

Interestingly, there is no subset method for JS's built-in sets, so we implemented our own (lines 615-629 in hyperschedule.js).

kwshi (approved):
Looks nice!

Issue #131, opened 2020-02-22
Credit Search Keyword


JeremyTsaii:
User Story:

As a user, I want to filter classes by course credits so I can create a schedule with the number of units I want. There have been scenarios where I want to search for 1 credit classes since I am at the cusp of overloading (16 or 17 credits), but there has been no way of looking for such courses, even on Portal. Having such a feature would allow for me to easily find classes with a specified amount of credits and help in crafting my schedule.

Implementation:

- [x] Add credit keyword

The user will be able to enter a query like “1 credit” and classes that are 1 credit will show up in the search results. We also want the user to be able to input other keywords before/after the time/day keywords to further narrow their searches.

![credit keyword](https://github.com/JeremyTsaii/Images/blob/master/credit.JPG)

This is part of CS121 Group 2. Team members include @JeremyTsaii, @godpng, @rory6103.

kwshi:
This sounds like a great idea that Hyperschedule can readily use.  

In the interest of consistency and simplicity, I'm inclined to suggest that we do so using Hyperschedule's existing GitHub-style "keyword" filtering mechanism (you can read more about it by clicking the "How does it work?" button in the Hyperschedule UI), instead of fuzzy-matching for search strings containing "n credits".  This means we would filter on credits not by querying things like "1 credit" or "2 credits" but rather adding keywords like "credits:1" and "credits:3", etc.  

Also worth considering are more sophisticated filtering like "credits:<3" or "credits:<=3" to filter credits with inequalities.

If that sounds good to you, the relevant code to look at right now is the `filterKeywords` global and the `coursePassesTextFilters` function.

I can see a case made, however, for fuzzy syntax like "1 credit" because it's slightly more intuitive (maybe) to use and doesn't require knowing special syntax.  But I personally find that it would make parsing quite complicated and, past a certain point, unpredictable, kind of like Google search queries.

JeremyTsaii:
Thanks Kye.

Regarding using the keyword filtering mechanism, that is actually what we have done! Our pull request detailing the implementation will come in the next few days. We have also changed the help modal accordingly to show the new filter keywords.

The inequality filtering idea for credits is quite interesting and we will try to get that working.

Thanks for the suggestions and feedback.

PR #135, opened 2020-02-29
Implement "credits:" Filter (Issue #131)

rory6103:
Implements a new course search keyword filter "credits:" that allows users to filter courses by the credits of each course.  Users can now directly input credits or use inequality operators to filter courses (See @kwshi Kye's comment in Issue #131).
credits-coursearch

The help modal was also changed to explain the new feature:
credits-helpmodal


kwshi:
I like it!  However, I won't merge this until it passes the checks--the good news is the only thing that needs fixing is the formatting (afaik).  Have you tried running `make format`?

jacobyoo47:
Thanks for the feedback @kwshi ! 

Looks like we forgot to run "make format" again before committing (oops). Hopefully all the checks pass now.

kwshi (changes requested):
Two comments--

It seems like there's an extra `hyperschedule.js` hanging out at the root level?  Did you accidentally make a copy of it and commit it?  I think we should remove it.

I made a small change to the documentation/help for the `credits:` filter in `index.html`, namely I changed one of the examples to be `<2` to cover more cases, etc.

src/js/hyperschedule.js
```diff
@@ -536,16 +539,55 @@ function coursePassesTextFilters(course, textFilters) {
...
+    (textFilters["college:"] && !col.match(textFilters["college:"])) ||
+    (textFilters["credits:"] &&
+      !courseCreditMatch(credits, textFilters["credits:"]))
   ) {
     return false;
   }
   return true;
 }
 
+function courseCreditMatch(courseCredits, inputCredits) {
```
kwshi: I suggest splitting out the parsing logic into a separate function, partly because I (stylistically) prefer seeing more `const`s than `let`s, and for efficiency reasons.  Below is some suggested code:

```js
function parseCreditsInequality(inputCredits) {
  for (const rel of filterInequalities)
    if (inputCredits.startsWith(rel)) 
      return rel;
  return "";
}

function parseCreditsFilter(inputCredits) {
  const rel = parseCreditsInequality(inputCredits);
  return { rel, floatInput: parseFloat(inputCredits.substring(rel.length) };
} 

function courseCreditMatch(courseCredits, inputCredits) {
  const { rel, floatInput } = parseCreditsFilter(inputCredits);
  const floatCredits = parseFloat(courseCredits);
```

jacobyoo47 (approved):

jacobyoo47 (dismissed):

jacobyoo47:
Not entirely sure what was going on with the hyperschedule.js --I think something finicky may have happened with VSCode and autosave at some point. I went ahead and removed the extra file.

I also added the separate parsing logic. Thanks for the suggested code and feedback!

kwshi:
Hmm-- just had an idea of something that would be nice to have.  What if I want to compound filters?  That is, if I want courses with credits between 1 and 3, I could do `credits:>1 credits:<3`, but that doesn't currently seem to work--only the last credits filter is taken, and the other ones are not.  Maybe that's really a consequence of the way search filters currently are structured, so if it's too complicated we can postpone that enhancement to later.  But let me know what you think!

JeremyTsaii:
Hey Kye,

Instead of compounding the credit filter like "credits:>1 credits:<3" we were thinking that we would just do "credits:1-3" to represent the inequality of x credits where 1 <= x <= 3.

Doing it this way would prevent us having to change a lot of the current structure. What do you think?

kwshi:
> Hey Kye,
> 
> Instead of compounding the credit filter like "credits:>1 credits:<3" we were thinking that we would just do "credits:1-3" to represent the inequality of x credits where 1 <= x <= 3.
> 
> Doing it this way would prevent us having to change a lot of the current structure. What do you think?

Could work too--just make sure to document it in the "how does it work?" so users know how to use it.

JeremyTsaii:

JeremyTsaii:
Hey Kye,

Conversion to TypeScript is done. Moving on to the next!

mia1024:
Closed due to to inactivity

Issue #134, opened 2020-02-29
Professor Ratings


JeremyTsaii:
User Story:

As a user, I want to view professor ratings from rate-my-professor directly on Hyperschedule so that comparing professors for the same course is more convenient. This would allow me to not have to manually search up professors through various tabs. Although ratings are not always indicative of a professor's quality, it provides a rough idea of what other students think of them and provides a tool for ease-of-access.

Implementation:

There are 2 ways we have in mind for implementing this. 

1. Simple Link (easier, less aesthetic, still quite convenient):

Our end goal is to make the row containing the professor's name on Hyperschedule become a clickable link that will take them to the appropriate RMP page. Rate my professor does not have a public API so we have to go around some obstacles to get the link to a given professor's page. Note that we would not have links for PE classes and classes being taught by Staff. First, a search query request is sent to RMP with the professor's name and college (taken from course information rows on Hyperschedule). This request will have the form "https://www.ratemyprofessors.com/search.jsp?queryBy=teacherName&schoolName=' + campus_name + '&queryoption=HEADER&query=' + teacher_name + '&facetSearch=true'". The request will come back with JSON showing the results of the search. There are several cases that could happen:

- No search results
This means the combination of the professor name and school do not exist on RMP. In this case, we would make the link direct to a page that has a search query of ONLY the professor's name (perhaps the professor has a page under a different college since they are new to the 5C's). The user will then be able to click on the professor's name on Hyperschedule and be directed to an RMP page showing professors with the same name and they will scroll through and find the corresponding professor. The convenience in this case is that users will not have to manually search for the professor in an alternate tab.

- One search result
This is the optimal case. We would scrape the underlying URL of the single search result and direct the link of the professor's name to this URL. The user would then be able to click on the professor's name on Hyperschedule and be instantly sent to the corresponding RMP page with ratings and reviews. No need to do a manual search and scroll through the results for the right professor. Note that sometimes professors for a specific school will have pages on RMP but they will have no ratings. In this case, we could make the URL like in the no search results case, simply searching the professor's name on RMP to see if they have a page with actual ratings under a different school.

- Multiple search results
Since we are matching professor name and college, we will just take the first search result and treat it like the one search result case. Possible issues are if there are two professors with the same name at the same college, but perhaps we could add a third search query with the college department (this is possible on RMP). 

2. Statistics Popup (harder, possible legal issues, visually appealing):

[...]

PR #141, opened 2020-04-15
Implement RateMyProfessor Links (Issue #134)

JeremyTsaii:
Implements the ability for users to click on professor names in the course description box and have a new tab open to RateMyProfessors with the professor's name as the search query. In future implementations, could possible advance this by making the link go to the professor's actual RMP page or have a popup display the professor's data without having to go to a new tab.

image

image

Also updates the help modal to inform users of where the links take them and how they should understand RMP takes anonymous reviews.

image


JeremyTsaii:
Kye, we were not sure if this feature was wanted. What do you think? I think it will be helpful personally.

Issue #140, opened 2020-04-15
Multiple Schedules Functionality


JeremyTsaii:
User Story: As a user, I want to be able to draft multiple schedules and have Hyperschedule save them. This is extremely useful for choosing alternate courses for a single semester and being able to compare the different schedule options available to you. Currently, without this functionality, I find that I must manually star/unstar courses when looking at alternates and the selected course list also expands and gets cluttered. 

We plan to allow users to have 4 separate schedules which can be switched around by simply clicking on one of 4 buttons denoting the specific schedule (look at diagram below). These buttons will be visible on the course search tab. We also want the users to be able to name these 4 schedules so that customizability and organization are increased. Clicking the plus button when adding a course will then add the course to the schedules that are currently selected and not the other ones. 

When viewing the schedule tab, there will be 4 buttons to view the specific schedule on the calendar interface. 

Diagram of our intended implementation:

![Hyperschedule Issues](https://user-images.githubusercontent.com/44514622/79398022-779c9980-7f34-11ea-83c4-d6c221438039.jpg)

This is part of CS121 Group 2. Team members include @JeremyTsaii, @godpng, @rory6103.

mia1024:
Implemented in the new version

PR #143, opened 2020-04-24
Multiple Schedules (Issue #140)

JeremyTsaii:
Implements the ability for users to keep track of 4 separate schedules and add courses to them individually. For even easier use, users are able to add classes to multiple schedules simultaneously (adding a course adds to all schedules with a check by their name). In the future, we may have the option of renaming schedules (right now they are Schedule1-Schedule4). 

Screenshot from 2020-04-23 22-13-54

Also updated help modal to reflect these changes.

Screenshot from 2020-04-23 22-20-14

Screenshot from 2020-04-23 22-20-27

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -75,11 +75,12 @@ body {
 }
 
 #course-search-schedule-toggle-bar {
-  margin: 1rem 0 1rem 1rem;
+  margin: 1rem 0 0 1rem;
 }
 
 #course-schedules-buttons-bar {
   margin: 1rem 0 1rem 1rem;
+  padding: 0 0;
   justify-content: space-around;
 }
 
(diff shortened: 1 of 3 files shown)


kwshi:
This is headed to a really nice feature that we've been wanting since forever, but I feel like a fixed set of four schedules isn't exactly the way to go.  I think it makes more sense to start from a single schedule, and have "create new"/"delete"/"edit" options on a list of schedules.  Some people don't want multiple schedules--having four checkboxes as tabs might be confusing/distracting; others might want five, or six, or tens--four would be limiting.  What do you think?

JeremyTsaii:
Hey Kye, thanks for your suggestions. We were thinking that most people would never need more than 4, but you do have a point that there are potential outliers to this.

Do you propose that we have a drop down menu to select the current schedule and also have a drop down menu to check off schedules that you would want the courses to be added to (instead of having the schedule tabs across the screen)?

PR #146, opened 2020-05-02
Add Multiple Schedules (Issue #140)

JeremyTsaii:
This is the reworked version of the previous PR regarding multiple schedules. Now there is a dropdown that allows users to create as many schedules as they want. Users can rename and delete the schedules that they make. Bonus features include colorful schedule rows, color changing schedule tab/dropdown, and the schedule tab showing you the name of your currently-selected schedule!

Screenshot from 2020-05-01 21-53-27

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -130,17 +130,10 @@ body {
 
 /** Multiple schedules buttons **/
 
-.course-schedule-button-content {
-  display: flex;
-  flex-direction: row;
-  justify-content: center;
-}
-
 .schedule-checkbox {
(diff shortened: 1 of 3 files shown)


JeremyTsaii:
Update: Will change this PR to TypeScript and resolve any merge conflicts. Since this PR is larger, it will take some more time.
Kye, did you want to keep the random coloring of schedules or should we just have them be uniform? I think the coloring looks cool but also makes the code a bit more jank with the bootstrap overrides.

mia1024:
Closed due to inactivity

PR #147, opened 2020-05-15
Show Only Hums/Techs Filter

JeremyTsaii:
Implements a hum and tech course filter. Often times, we are looking for only hums or only techs that satisfy certain search keywords (such as a specific college or time frame). One specific example is looking for Mudd hums specifically. While we can currently filter by college, we can only specify a single department. With this implementation, this is possible by simply checking the "Show humanities courses (only)" checkbox in the filters drop down and then specifying "col:HM" in the course search area.

Screenshot from 2020-05-15 16-27-54


JeremyTsaii:

JeremyTsaii:
After combing through literally every single course in the hyperschedule api, I've added CSMT, ASTR, DS, GEOL, and NEUR codes to the tech filter list. These are some of the more obscure codes that fall under the classification of a more STEM/technical course.

src/js/course.ts
```diff
@@ -470,6 +485,16 @@ export function isClosed(c: CourseV3) {
...
+export function isTech(c: CourseV3) {
+  const code = c.courseCode.split(" ")[0];
+  return filterTechs.includes(code) && code !== "PE";
+}
+
+export function isHum(c: CourseV3) {
+  const code = c.courseCode.split(" ")[0];
+  return !filterTechs.includes(code) && code !== "PE";
+}
+
```
kwshi: Hmm, I think a more TypeScript-y, and also future-proof, way to implement this sort of categorization is via enums.  I suggest declaring an enum type (named `Area` or `Techiness` or something you can come up with that's better) with entries `Tech`, `Hum`, and `Pe`, and having a function `area(c: CourseV3): Area` instead of separate `isTech`, `isHum` functions and having to check `!== "PE"` in each of them.

JeremyTsaii: That's a great organizational idea. I will go ahead and do that soon!

src/js/course.ts
```diff
@@ -470,6 +491,17 @@ export function isClosed(c: CourseV3) {
   return c.courseEnrollmentStatus === "closed";
 }
 
+export function getCourseArea(c: CourseV3): CourseArea {
+  const code = c.courseCode.split(" ")[0];
```
kwshi: I actually think I'd prefer using regexp here, since there's technically no API-side guarantee that the course code contains a space between the department and number (e.g. `CSCI 005` instead of `CSCI005`).  Technically, it doesn't guarantee that course codes begin with their department at all, but using a regexp like `/^[A-Z]+/` would be a little more forgiving than simply splitting on a space.  Ideally, what we would want to do is add a `department` field to the API response spec.

src/js/course.ts
```diff
@@ -470,6 +491,17 @@ export function isClosed(c: CourseV3) {
...
+  if (code === "PE") {
+    return CourseArea.Pe;
+  } else if (filterTechs.includes(code)) {
+    return CourseArea.Tech;
+  } else {
+    return CourseArea.Hum;
+  }
```
kwshi: This is short enough that I'm inclined to prefer a conditional/tertiary _expression_ here instead:
```typescript
code === 'PE' ? CourseArea.Pe
  : filterTechs.includes(code) ? CourseArea.Tech : CourseArea.Hum;
```

src/js/hyperschedule.ts
```diff
@@ -1263,6 +1288,22 @@ function toggleClosedCourses() {
   writeStateToLocalStorage();
 }
 
+function toggleTechCourses() {
+  gShowTechCourses = !gShowTechCourses;
+  if (gShowTechCourses) gShowHumCourses = false;
```
kwshi:
```typescript
gShowHumCourses = gShowHumCourses && !gShowTechCourses;
```

They say cool kids do branchless programming, and I wanna be cool :nerd_face:

src/js/hyperschedule.ts
```diff
@@ -1263,6 +1288,22 @@ function toggleClosedCourses() {
...
+  updateShowHumCoursesCheckbox();
+  updateCourseSearchResults();
+  writeStateToLocalStorage();
+}
+
+function toggleHumCourses() {
+  gShowHumCourses = !gShowHumCourses;
+  if (gShowHumCourses) gShowTechCourses = false;
```
kwshi:
```ts
gShowTechCourses = gShowTechCourses && !gShowHumCourses;
```

kwshi:
On another note, you may (?) get a slight performance gain from using `const filterTechs = new Set([...])` instead of `[...]` and using `.has` instead of `.includes` to check whether courses are techs/hums/PE.

JeremyTsaii:
> On another note, you may (?) get a slight performance gain from using `const filterTechs = new Set([...])` instead of `[...]` and using `.has` instead of `.includes` to check whether courses are techs/hums/PE.

I agree. Gotta get that O(1) check :)

src/js/course.ts
JeremyTsaii: Makes sense. Changed to regexp matching instead.

stuxf:
closing, implemented in new versions of hyperschedule with the course area filter

PR #150, opened 2020-08-07
Change schedule to start at 7am to accomodate 7:30am classes fall 2020

ingridt123:


kwshi:
First off, let me say thank you so much for doing this!  I know Hyperschedule has been having all sorts of issues lately, and I've been too busy to get around to fixing them quite yet.  So thanks for the help.

The deploy preview given by Netlify suggests that this doesn't seem to be working correctly--the JS console is logging a bunch of "`timeSince8am` is not defined" errors.  Can you check to see what's going on?

kwshi:
This looks solid; great work, and thanks again for this fix!

kwshi:
I'm an idiot and accidentally closed this instead of merging it.

PR #151, opened 2020-08-09
Add time zone toggle to configure schedule in different time zones

ingridt123:
Creates dropdown in settings to select from all time zones (which changes all below based on dropdown selection)
- Course schedule times and days
- Course description times and days
- Print course schedule times and days
- Import/export JSON code courses times and days

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -420,6 +420,96 @@
                 Gray out courses conflicting with any course</label
               >
             </div>
+            <br>
+            <div id="time-zone-dropdown-wrapper">
+              <label id="time-zone-dropdown-heading"
+                >Choose time zone to display schedule</label
+              >
+              <select id="time-zone-dropdown">
+                <option timeZoneId="1" gmtAdjustment="GMT-12:00" useDaylightTime="0" value="-12.0">(GMT-12:00) International Date Line West</option>
+                <option timeZoneId="2" gmtAdjustment="GMT-11:00" useDaylightTime="0" value="-11.0">(GMT-11:00) Midway Island, Samoa</option>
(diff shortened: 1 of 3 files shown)


kwshi:
It seems like there are a few small issues--

- Upon refreshing the page, the selected time _zone_ stays the same, but the location associated with it may not.  For instance, if I select "eastern time (GMT-5)", refreshing gives "Bogota, Lima, Quito, Rio Branco (GMT-5)".  This is slightly eyebrow-raising, though it does not affect the core functionality.
- Is the time zone also exported in the JSON?  Or are the timezone-adjusted times for the courses exported directly?  What if I
  1. set my time zone to something non-default, like eastern time, then
  2. export JSON, then
  3. start a new session (say, incognito, or another computer) _without_ changing the time zone (i.e. use the default), and then
  4. import the JSON--
  Will the courses be imported with the correct times?  Or will they be imported with the pre-adjusted times, but then in the wrong time zone?

Apart from those two things, this looks great!  Thanks so much for the work!

kwshi:
I'd like to make a recommendation to always _store the courses in the same "standard" time zone_ (probably PST), and only adjust for time zone in the rendering steps (adjust the HTML labels).  That would save us from having to worry about time zones in the exported JSON course data, since all TZ conversions only happen on the "visual" side.

ingridt123:
Yeah I agree I think that is probably the most clear/easy. On another note, I was thinking that it might be nice to add time zone for the iCal export since it currently exports the times as "floating" (so whatever time zone a user's computer is configured to, a 9:15am PST would be 9:15am in their time zone).

ingridt123:
Just FYI all the comments should be addressed. I can do the iCal change in a separate PR. Thanks!

Update: I've skimmed through the ics code and I don't think configuring time zone has been implemented yet.

kwshi:
Nice!  This is great.  Thanks!

PR #153, opened 2020-08-16
Fix problems due to daylight savings

ingridt123:
Accounts for daylight savings around the world while toggling time zone


kwshi:
Nifty!

PR #154, opened 2020-08-23
Make Timezone Changes Persist on Refresh

nguyenank:
Before, the selected timezone persisted on refresh in the dropdown, but the actual schedule/course timings were still in Pacific time. I just made it so when the selected timezone is set on startup, it also fires an event as if the timezone has just been changed via dropdown to adjust all the course timings on startup.


kwshi:
Good catch, nifty fix.  Thanks!

PR #155, opened 2020-09-29
Setup Typescript, Webpack

kwshi:
This PR converts our vanilla JS codebase to Typescript and switches the build system from Parcel to Webpack.  It also drops the Makefile/Docker infrastructure and relies purely on `npm` for builds.

The rationale for these two decisions:
- Using static typing offers much stronger guarantees for safe code at compile-time, catching errors such as calling functions with the wrong number of parameters, wrong types, mistyped names, possibly-undefined values, etc.  It also provides extra confidence that we are calling functions correctly (not swapping their arguments), especially in the event that our libraries' APIs change.  Case in point: while I was making this conversion, I found in our _existing_ code a handful of these errors: passing a number to `parseInt`, an unused `slot` parameter in `createSlotEntities`, mutating a `const` value, etc.  Of various JS static-typing solutions, Typescript is the most popular/well-established/well-supported.
- Parcel technically "supports" Typescript, but its support is very limited (see [this discussion](https://github.com/parcel-bundler/parcel/issues/4022)): it ignores type annotations and simply transpiles them using Babel (which also [doesn't actually support the full Typescript spec](https://kulshekhar.github.io/ts-jest/user/babel7-or-ts)) without performing any typechecking.  (Hence it also ignores Typescript's `tsconfig.json` configuration, providing no customization capabilities whatsoever.)  There is allegedly a Parcel plugin for improving Typescript support, but I cannot find it, and it likely does not have much official support, because the Parcel authors have a weird stance on what they think Typescript is for.  Anyway, so Webpack offers out-of-the-box support for Typescript and is much more configurable, which makes it a more viable long-term solution for TS projects.
- Docker's main purpose is to enable reproducible/deterministic builds by packaging all library/tooling-dependencies into a Docker container.  In the case of Hyperschedule, the only software needed to build its frontend is `npm` (along with JS packages installed via `npm install`), which is easily found/installed on every major OS and comes pre-packaged with Netlify's build container).  Having only one dependency renders Docker somewhat overkill.  Instead, simply having `npm` installed suffices; `npm`'s own `package_lock.json` version locking mechanism means that, as far as building the front-end go, everything remains deterministic.  Similarly, `npm`'s `package.json` provides a `scripts` mechanism that, while slightly less versatile than a Makefile, is easier to work with, simpler, and totally sufficient for our needs.  Removing Docker and Make from the build pipeline vastly simplifies things.

Current caveats:
- The types are not yet sound: many types are not yet checked and implicitly ignored or marked `any`, in the interest of quickly migrating to Typescript first and getting things to compile _and then_ fixing errors.  A `"strict": true` option needs to be added to `tsconfig.json`, and all resulting type errors need to be fixed.
- I am not sure whether I've done the Netlify configuration right, so, uh, we'll see!
- Also, I haven't enabled/configured analytics correctly yet.


kwshi:
Well, the Netlify deploy preview seems to be working fine, so I think that's not an issue.  I'm going to do some more cleaning up before merging this.

kwshi:
I've reorganized a bunch of things into modules so the code is easier to navigate, enabled "strict" mode on TypeScript (so the type-checking is much more sound), and restored feature parity with current Hyperschedule (with a few minor caveats).  It's still not perfect, but I'm merging sooner rather than later so that incoming PRs won't have to do a lot of rewriting.

PR #161, opened 2020-11-04
Changes to selected courses bar to access all buttons

ingridt123:
Currently the selected courses bar is not formatted well and for smaller screens doesn't show some of the buttons:

Screenshot 2020-11-04 at 1 37 01 PM

<br/>

Thus removed the text from the GitHub link (shows "View on GitHub" tooltip if you hover over it) and added horizontal scroll so that the user can access all the buttons even on smaller screens.

Screenshot 2020-11-04 at 1 37 19 PM
Screenshot 2020-11-04 at 1 38 10 PM

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -250,7 +250,12 @@ body {
 
 #selected-courses-bar {
   overflow-y: auto;
-  justify-content: space-evenly;
+  justify-content: space-between;
+}
+
+#selected-courses-bar::before,
+#selected-courses-bar::after {
+  content: "";
 }
 
 #github-link-wrapper {
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -249,10 +249,17 @@ body {
 /*** Selected courses header bar ***/
 
 #selected-courses-bar {
+  overflow-y: auto;
+}
+
+#github-link-wrapper {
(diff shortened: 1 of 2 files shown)


kwshi:
Neat.  I wonder if it'll look better with different spacing/alignment?  I suggest playing with the `#selected-courses-bar` element's `justify-content` CSS attribute (try `flex-end` or `space-between`), or try setting the Github button to `flex-grow: 1`.  What do you think looks best?

ingridt123:
Thanks for the comment. I experimented with a few options and I think space-evenly looks the best (the second-bottom-most screenshot). Let me know what you think!

UPDATE: space-evenly looks best but it doesn't work with the scrolling, so updated to use space-between with added spaces before and after (the bottom-most screenshot).

Original
Screenshot 2020-11-06 at 12 31 24 PM

`justify-content: flex-end`
Screenshot 2020-11-06 at 12 31 15 PM

`justify-content: space-between`
Screenshot 2020-11-06 at 12 31 01 PM

`justify-content: space-evenly`
Screenshot 2020-11-06 at 12 47 44 PM

`justify-content: space-between` with added content before/after
Screenshot 2020-11-06 at 1 25 40 PM

PR #162, opened 2020-11-06
Remove global scroll

ingridt123:
Scroll within the course schedule and selected courses list instead of global scroll to make it more readable/easier to navigate


kwshi:
Nice fix!  Can I suggest simply adding CSS attributes to `.columns` for now without creating a new `#wrapper` ID, and using `min-height: 0` instead of `overflow: hidden` (which, I think, achieve the same effect, but `min-height: 0` does so more explicitly)?

PR #165, opened 2020-11-13
Fix selected courses toolbar height issue

ingridt123:
I've found that in some situations the height of the selected courses toolbar shrinks, so this fixes the problem.

Screenshot 2020-11-12 at 11 56 10 AM

Screenshot 2020-11-12 at 10 43 49 PM

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -249,6 +249,7 @@ body {
 #selected-courses-bar {
   overflow-y: auto;
   justify-content: space-between;
+  min-height: 38px;
 }
 
 #selected-courses-bar::before,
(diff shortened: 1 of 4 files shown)


src/css/main.css
```diff
@@ -249,6 +249,7 @@ body {
 #selected-courses-bar {
   overflow-y: auto;
   justify-content: space-between;
+  min-height: 38px;
```
kwshi, line 252: Hmm, hard-coding pixel sizes is not really great CSS--it becomes inconsistent depending on available fonts, different devices/DPIs, etc.  Have you tried `min-content` or `max-content` instead, since the goal is to fit the container to its content?

kwshi (changes requested):

kwshi:
I recently redid a handful of the styling; this should not be needed anymore.

PR #166, opened 2020-12-01
Implement Time Filter

JeremyTsaii:
(Closed and Reopened from last year)

Implements a new course search filter for time that allows users to filter courses by adding a starting time (such as 11:00am) OR a range of times (11:00-3:00pm), where the range of times filters classes that start and end within the time range.

This completes the conversion of past PRs revolving around filters to TS (hum/tech, time, credits). My next goal is to convert the multiple schedules PR to TS, after these PRs are approved. Thanks!


src/js/hyperschedule.ts
```diff
@@ -527,9 +527,51 @@ function coursePassesTextFilters(
...
 
-function parseDaysInequality(inputDays: string) {
-  for (const rel of filterInequalities)
-    if (inputDays.startsWith(rel)) return rel;
+// timeFilters is a one or two element array
+// [start_time] or [start_time, end_time]
+function coursePassesTimeFilters(
+  course: Course.CourseV3,
+  timeFilters: string[]
+) {
```
kwshi: This is better described by the type `[string] | [string, string]`.

src/js/hyperschedule.ts
```diff
@@ -527,9 +527,51 @@ function coursePassesTextFilters(
...
+  // indicates no current time filters
+  if (timeFilters[0] === "") {
+    return true;
```
kwshi: Is there a way to filter on _just the end time_/is that something you should be able to do?  If so, maybe this check should also consider `timeFilters[1]`.

src/js/hyperschedule.ts
```diff
@@ -527,9 +527,51 @@ function coursePassesTextFilters(
...
+  }
+
+  const startTime = parseFloat(timeFilters[0].replace(":", "."));
```
kwshi: Eh?  This seems kind of like a strange way to parse times--consider the following inputs:
- `5:3`--should this be either considered invalid, or it should parse as `5:03`, but currently it would be equivalent to `5:30`.
- `4:45pm`--`parseFloat` will simply ignore the `pm`, contradicting the example usage provided by the "help" popup.  Or am I overlooking some other code that does some pre-processing here?

Also, maybe all this new filtering code should be placed in a separate, new, `filter.ts` module, so as to avoid cluttering up `hyperschedule.ts` even more.  I'm in a slow ongoing process to split up `hyperschedule.ts` into smaller modules; filters are a good place to start.

src/js/hyperschedule.ts
```diff
@@ -527,9 +527,51 @@ function coursePassesTextFilters(
...
+  // if multiple start times, specified start must match at least one
+  if (timeFilters.length === 1) {
+    for (const schedule of course.courseSchedule) {
+      let scheduleStart = schedule.scheduleStartTime.replace(":", ".");
+      if (startTime === parseFloat(scheduleStart)) {
```
kwshi: Do these allow for inequality queries (e.g. `>9:35`)?  It doesn't seem apparent to me that it does.

src/js/hyperschedule.ts
kwshi: Another potential issue with using `parseFloat` is that floating-points are more susceptible to rounding errors, possibly, which may lead to, _maybe_, in some cases, `9:35` becoming `9.35000001` or something and not matching a `9:35` class.  It doesn't seem very likely, but maybe it's good to avoid it if we can, anyway.

src/js/hyperschedule.ts
kwshi: Either this should be fixed, or the documentation in the "help" popup should be corrected.

src/js/hyperschedule.ts
```diff
@@ -854,15 +897,86 @@ function processSearchText() {
...
+  // Ex: 3:00PM or 3:00PM-4:00PM
+  return (
+    (timeArray.length === 3 &&
+      !isNaN(+timeArray[0]) &&
+      timeArray[1].includes("-")) ||
+    (timeArray.length === 2 &&
+      !isNaN(+timeArray[0].charAt(timeArray[0].length - 1)) &&
+      !isNaN(+timeArray[1].substring(0, 2)))
+  );
+}
```
kwshi: This stuff seems kind of convoluted.  In fact, it doesn't appear to be used by `getTimeRange` or `getTimeFilter`, so is it needed at all?

src/js/hyperschedule.ts
```diff
@@ -854,15 +897,86 @@ function processSearchText() {
...
+  }
+  let timeArr = [];
+  for (let time of inputArr) {
+    if (time.substring(time.length - 2) === "am") {
+      // remove "am"
+      time = time.substring(0, time.length - 2);
+      if (time.startsWith("12")) {
+        time = "00:".concat(time.substring(3));
+      }
+    } else if (time.substring(time.length - 2) === "pm") {
```
kwshi:
```typescript
const pre = time.slice(0, time.length-2);
const suf = time.slice(time.length-2);
const [hour, min] = time.split(':');
if (suf === 'am') {...}
else if (suf === 'pm') {...}
else {...}
```
DRY

src/js/hyperschedule.ts
```diff
@@ -854,15 +897,86 @@ function processSearchText() {
...
+      // remove "pm"
+      time = time.substring(0, time.length - 2);
+      let [hour, min] = time.split(":");
+      if (!time.startsWith("12")) {
+        let hourNum = parseInt(hour) + 12;
```
kwshi: `parseInt` should always be called with `parseInt(n, 10)`; otherwise it tries to do some funky thing where it tries to interpret numbers in other bases, etc.

Thanks, Javascript!

kwshi (changes requested):

src/js/hyperschedule.ts
JeremyTsaii: This is currently used by `processSearchText()` to see if a search query is indeed a specified time/time range. I believe the `!isNan` was an interesting way we found to check if a string is numeric, but I believe this can be done better through a regex.

src/js/hyperschedule.ts
JeremyTsaii: The inequalities are actually handled in `getTimeRange`. Depending on the inequality, the time filter pushes an upper or lower bound of "00:00" or "23:59". Hence, the only case when the time filters length is equal to 1 is when no inequality or "=" was passed. That's why we check "===" for the single case.

src/js/hyperschedule.ts
JeremyTsaii: Good catch! I noticed there are some other places in the code where parseInt is called without base 10. Just btw.

src/js/hyperschedule.ts
JeremyTsaii: The elements within the time filters are already processed through `getTimeFilter` so that means all elements are between 00:00 and 23:59, without "am"/"pm". I believe we used `parseFloat` for the second case where there could be a range of times and thus checking if a course's start time is between the range of times is easier with floats. For the first case, string comparison would be enough. However, the second case is trickier with the range. There is a way with `parseInt` instead, where we can convert the times to number of minutes and then compare that. Does that sound like a good way to you (gets past `parseFloat` at least)?

src/js/hyperschedule.ts
JeremyTsaii: Right now, there is no way to filter on the end time. With the way the regex is set up now, there isn't a way for there to be a `timeFilters[1]` if there isn't a `timeFilters[0]`. I personally think filtering on end time might be a bit weird.

JeremyTsaii:
Okay, I've gone ahead and modularized the filter functions into `filter.ts`. If you have time, could you also take a look at the Credits Filter (#135)? Thanks for your comprehensive reviews!

mia1024:
Closed due to inactivity

PR #180, opened 2021-11-05
svelte :)

kwshi:
- start migrating to svelte
- configure to build w/ svelte-kit


stuxf:
"react is better, sorry" - mia

PR #183, opened 2021-11-20
Turn links in course descriptions into anchor tags in the DOM

marcos-acosta:
**Description**
I noticed that links appear as literal anchor tags in course descriptions, especially for PE classes. I figured it would be convenient for users if those links were displayed as actual anchor tags.

**Changes**
- Add CSS for anchor tag in `course-details.pcss`
- Add a function for using regex to parse raw anchor tags in course descriptions
- Set description with `.innerHTML()` instead of `.textContent()`
  - `innerHTML` is less efficient than `textContent`, but I imagine the performance drop is marginal

### DIFF
--- a/src/css/course-details.pcss
+++ b/src/css/course-details.pcss
@@ -7,6 +7,11 @@
     @apply pb-4;
   }
 
+  a {
+    color: rgb(59, 130, 246);
+    text-decoration: underline;
+  }
+
   header {
     @apply font-bold grid;
     grid-template-areas: "code close" "name close";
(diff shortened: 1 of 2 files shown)


kwshi:
Nice idea!  Can I ask-- why is innerHTML needed if you're already making a regexp parser for anchor tags?  Or vice versa?

marcos-acosta:
Thanks! You were right, it doesn't need the regex parser, setting `innerHTML` to the raw course description displays links properly too

kwshi:
By the way, can you tell me an example of such a course (with anchor tag in description) so that I can test out & compare?

marcos-acosta:
Yeah! Almost all of the PEs have a link to the CMS athletics webpage, e.g. ping pong

kwshi:
Looks nice!

kwshi (approved):

PR #188, opened 2023-01-30
Add functions to convert section data from APIv4 to V3 

foobles:


shared/api/v3/course.ts
```diff
@@ -0,0 +1,160 @@
...
+export interface Schedule {
+    scheduleDays: string;
+    scheduleStartTime: string;
+    scheduleEndTime: string;
+    scheduleStartDate: string;
+    scheduleEndDate: string;
+    scheduleTermCount: integer;
+    scheduleTerms: integer[];
+    scheduleLocation: string;
+}
```
kwshi: it would make my life easier if you defined these types in [zod](https://github.com/colinhacks/zod):

```typescript
import * as Z from 'zod';

const Course = Z.object({
  courseCode: Z.string(),
  courseName: Z.string(),
  // ...
  courseCredits: Z.number(),
  courseSeatsTotal: Z.number().int(),
})
export type Course = Z.TypeOf<typeof Course>

const Schedule = Z.object({
  // ...
})
export type Schedule = Z.TypeOf<typeof Schedule>
```

since eventually we're going to need all API-served types defined in Zod anyway, for type-safety/validation.

shared/api/v3/course.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    if (!Number.isInteger(total_minutes)) {
+        throw Error(`Time too precise to fit in hours:minutes format: ${secs}`);
+    }
+
+    const hours = Math.floor(total_minutes / 60);
+    const minutes = total_minutes % 60;
+
+    const hours_string = hours.toString().padStart(2, "0");
+    const minutes_string = minutes.toString().padStart(2, "0");
```
kwshi: generally TS/JS code style prefers `camelCase` to `snake_case`; certainly we should do that here for consistency with all the other variable naming conventions in the existing codebase

shared/api/v3/course.ts
```diff
@@ -0,0 +1,160 @@
...
+} {
+    return {
+        terms: [],
+        courses: sections.map(courseFromV4Section),
+    };
+}
+
+function stringifyWeekdayEnumArray(arr: APIv4.WeekdayEnum[]): string {
+    return arr.join("");
+}
```
kwshi: it doesn't really feel like this is complex enough to warrant an actual function, I kind of feel like you could just inline it

shared/tests/APIv3.test.ts
```diff
@@ -0,0 +1,149 @@
...
+function compareSchedules(x: APIv3.Schedule, y: APIv3.Schedule): number {
+    const xString = JSON.stringify(x);
+    const yString = JSON.stringify(y);
+    if (xString < yString) {
+        return -1;
+    } else if (xString > yString) {
+        return +1;
+    } else {
+        return 0;
+    }
```
kwshi:
```typescript
return JSON.stringify(x).localeCompare(JSON.stringify(y))
```

foobles: Thanks! I am gonna keep `stringifyWeekdayEnumArray` as a separate function just so the code is more readable, but I will redo everything else in correspondence with your suggestions.

shared/api/v3/course.ts
```diff
@@ -0,0 +1,160 @@
...
+            termCount: 2,
+            terms: [half.number - 1],
+        };
+    } else {
+        // Best guess given available data
+        return {
+            termCount: half.number,
+            terms: [half.number - 1],
+        };
+    }
```
kwshi: so, here's a style comment that maybe deviates from my (otherwise religious) devotion to FP-style conventions. when `if` blocks return early, don't put the remaining code inside `else` blocks:
```typescript
if (half === null) {
  return thing1;
} else if (half.prefix /* ... */) {
  return thing2;
} else {
  return thing3;
}

// instead, do
if (half === null) {
  return thing1;
}
if (half.prefix /* ... */) {
  return thing2;
}
return thing3;
```

there are two (subjective) advantages to this--
- reduces nesting.
- makes it clearer (? depends on who you ask, i guess) what branches of the logic are "short-circuit"/special cases, and which path corresponds to the "default"/catch-all path. though to be pedantic, in practice the "default" path here should be hit rarely/almost-never given the data we have/are working with, so application of this rule to this specific case may be taken with a grain of salt. my point remains as a general rule of thumb.

just to claim that this rule isn't _purely_ my own preference, but something that various other devs in the world online also care about-- it's sometimes called the "golden margin" rule, or the "align the happy path to the left edge" rule, articles/videos on those here:
- https://www.youtube.com/watch?v=CFRhGnuXG-4&t=216s
- https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

shared/api/v3/course.ts
mia1024: Adding on to that, we can also add numerical check to all the numbers. For example, we can set `courseSeatsTotal`, `permCount`, etc. to be `z.number().nonnegative()`. In fact, none of our numerical type should have negative numbers inside.

shared/api/v3/course.ts
```diff
@@ -0,0 +1,160 @@
...
+    };
+}
+
+function stringifyWeekdayEnumArray(arr: APIv4.WeekdayEnum[]): string {
+    return arr.join("");
+}
+
+function stringifyTime(secs: integer): string {
+    if (secs < 0 || secs >= 24 * 60 * 60) {
+        throw Error(`Time out of range: ${secs}`);
```
mia1024: Since your function can throw errors, be sure to add error handling to the function you will write to convert a list of v4 sections to v3 sections. If there is any error processing a section, log it and continue with the rest, and discard the invalid sections. We don't want crash the entire server just because we can't parse a singe course section.

shared/api/v3/course.ts
foobles: I am aware of this style I definitely agree with it. But like you said, the `else` case here is definitely a more anomalous situation, and doesn't represent the Happy Path, which is why I formatted it like this. I suppose I could change the conditions so that the Happy Path goes into the `else` case, and then I could take it out of the `if` altogether.

shared/api/v3/course.ts
```diff
@@ -0,0 +1,185 @@
...
+    courseName: z.string(),
+    courseSortKey: z.string().array(),
+    courseMutualExclusionKey: z.string().array(),
+    courseDescription: z.string(),
+    courseInstructors: z.string().array(),
+    courseTerm: z.string(),
+    courseSchedule: Schedule.array(),
+    courseCredits: z.number().nonnegative(),
+    courseSeatsTotal: z.number().int(),
+    courseSeatsFilled: z.number().int(),
```
mia1024: Both of them should be non-negative

shared/api/v3/course.ts
```diff
@@ -0,0 +1,185 @@
...
+            return schedule.locations.map((location) => ({
+                scheduleDays: days,
+                scheduleStartTime: startTime,
+                scheduleEndTime: endTime,
+                scheduleStartDate: startDate,
+                scheduleEndDate: endDate,
+                scheduleTermCount: termCount,
+                scheduleTerms: terms,
+                scheduleLocation: location,
+            }));
```
mia1024: You can inline all of the calls here to something like
```ts
return schedule.locations.map((location) => ({
                scheduleDays: schedule.days.join(""),
                scheduleStartTime: stringifyTime(schedule.startTime),
                scheduleEndTime: stringifyTime(schedule.endTime),
                scheduleStartDate: startDate,
                scheduleEndDate: endDate,
                scheduleTermCount: termCount,
                scheduleTerms: terms,
                scheduleLocation: location,
            }));
```

This increases both the readability of the code by not requiring the reader to jump back and forth between definitions and reducing runtime memory needed by not allocating extra once-off variables.

shared/api/v3/course.ts
```diff
@@ -0,0 +1,185 @@
...
+    courses: Course[];
+} {
+    return {
+        terms: [],
+        courses: sections.flatMap((section) => {
+            try {
+                const course = courseFromV4Section(section);
+                return [course];
+            } catch (e) {
+                console.error(e);
```
mia1024: After moving the functions to backend, you can use our logger by importing the `createLogger()` function from `logger.ts`. You can find examples of how it works by looking at files such as `hmc-api/course.ts` and `middleware.ts`. The library we are using is `pino` and its documentation might also help.

shared/api/v3/course.ts
```diff
@@ -0,0 +1,185 @@
...
+                return [];
+            }
+        }),
```
mia1024: If you change this chunk to a `for-of` loop you don't need to return empty arrays, but simply pushing data to a pre-allocated array. This avoids multiple allocations (and given the size of our data, about 2.5k) of arrays that are later merged back into the resulting array each time this function is called. You can just do `continue` inside the loop to skip pushing broken data.

shared/tests/APIv3.test.ts
```diff
@@ -0,0 +1,143 @@
+import * as APIv4 from "../api/v4";
```
mia1024: You can also move this file to the `tests` folder in backend once you moved the functions.

mia1024 (changes requested):
Can you move every single function in your current `course.ts` (that is, line 48 and below) into `backend` instead of `shared`? Only type definitions are used by both frontend and backend, and conversions like this will never happen on frontend. You can put everything under `hmc-api/v3-conversion.ts`.

shared/api/v3/course.ts
kwshi: it's up to you then, feel free to keep as is

shared/api/v3/course.ts
foobles: The reason I didn't inline these values is because I didn't want it to instantiate new strings for every single element of `location`, since `days`, `startTime`, and `endTime` would be the same for each one. By factoring them out, they all share a reference to the same string.

shared/tests/APIv3.test.ts
```diff
@@ -0,0 +1,135 @@
...
+import * as APIv3 from "../api/v3";
+
+import { describe, test } from "@jest/globals";
+
+const v4Section: APIv4.Section = APIv4.Section.parse({
```
mia1024: We don't need this

backend/src/hmc-api/v3convert.ts
```diff
@@ -0,0 +1,148 @@
...
+        logger.error({ seconds: secs }, "Time out of range");
+    }
+    const total_minutes = secs / 60;
+
+    if (!Number.isInteger(total_minutes)) {
+        logger.error(
+            { seconds: secs },
+            "Time too precise to fit in hours:minutes format",
+        );
+    }
```
mia1024: If this isn't integer, the resulted string from this function will be incorrectly formatted, and may cause more issues down the line. What we should do here is log the error, then either use `Math.round` or `Math.floor` to make sure the function always returns correctly formatted result.

backend/src/hmc-api/v3convert.ts
```diff
@@ -0,0 +1,148 @@
...
+}
+
+function stringifyWeekdayEnumArray(arr: APIv4.WeekdayEnum[]): string {
+    return arr.join("");
+}
+
+function stringifyTime(secs: number): string {
+    if (secs < 0 || secs >= 24 * 60 * 60) {
+        logger.error({ seconds: secs }, "Time out of range");
+    }
```
mia1024: if it's out of range, we should cap it to either 0 or 86400 in addition to logging the error to not produce broken data.

shared/tests/APIv3.test.ts
```diff
@@ -0,0 +1,68 @@
...
+            scheduleTermCount: 1,
+            scheduleTerms: [0],
+            scheduleLocation: "Beckman Hall B126",
+        },
+    ],
+    courseCredits: 3,
+    courseSeatsTotal: 30,
+    courseSeatsFilled: 28,
+    courseWaitlistLength: null,
+    courseEnrollmentStatus: "reopened" as APIv3.EnrollmentStatus,
```
mia1024, line 60: If you need to use type assertion here, it's better to do `courseEnrollmentStatus: APIv3.EnrollmentStatus.reopened`. The reason why we had type assertions at the end of some V4 sections is that it's the easiest way (alternatively, we would need to change about 5 different values). Inline assertion as an enum is slightly different from that.

mia1024 (approved):
Everything looks good, you can merge the PR after making the enum change.

PR #190, opened 2023-02-14
add Boomi database dump parser

foobles:
adds `parseCourseBoomi` and `parseBoomi` functions to parse the Boomi database dump format.

### DIFF
--- a/backend/src/hmc-api/files.ts
+++ b/backend/src/hmc-api/files.ts
@@ -72,10 +72,17 @@ export function parseCourseSectionSchedule(data: string) {
     return processResult(result, "courseSectionSchedule");
 }
 
+export interface ParsedCourse {
+    code: string;
+    title: string;
+    campus: string;
+    description: string;
+}
+
 /**
  * parser for course_1.csv
  */
-export function parseCourse(data: string) {
+export function parseCourse(data: string): ParsedCourse[] {
     const result = parse(
         {
             hasHeader: true,
@@ -96,28 +103,23 @@ export function parseCourse(data: string) {
     return courses;
 }
 
-export function parseCourseBoomi(data: string): {
-    code: string;
-    title: string;
-    campus: string;
-    description: string;
-}[] {
+export function parseCourseBoomi(data: string): ParsedCourse[] {
     const result = parseBoomi(
         ["code", "title", null, null, "campus", "description"],
         data,
     );
 
-    if (result.ok) {
-        for (const warning of result.warnings) {
-            logger.warn(warning, "Warning parsing Boomi database dump");
-        }
-        return result.records;
-    } else {
+    if (!result.ok) {
         logger.error(result.error, "Cannot parse Boomi database dump");
         throw Error(
             "Cannot parse Boomi database dump: " + JSON.stringify(result.error),
         );
     }
+
+    for (const warning of result.warnings) {
+        logger.warn(warning, "Warning parsing Boomi database dump");
+    }
+    return result.records;
 }
 
 /**
--- a/backend/src/hmc-api/files.ts
+++ b/backend/src/hmc-api/files.ts
@@ -1,4 +1,4 @@
-import { parse } from "./parser";
+import { parse, parseBoomi } from "./parser";
 import type { Result } from "./parser";
 import { z } from "zod";
 import { createLogger } from "../logger";
@@ -94,6 +94,30 @@ export function parseCourse(data: string) {
         c.description = c.description.replaceAll("||``||", "\n");
     }
     return courses;
+}
+
+export function parseCourseBoomi(data: string): {
+    code: string;
+    title: string;
+    campus: string;
+    description: string;
+}[] {
+    const result = parseBoomi(
+        ["code", "title", null, null, "campus", "description"],
+        data,
+    );
+
+    if (result.ok) {
+        for (const warning of result.warnings) {
+            logger.warn(warning, "Warning parsing Boomi database dump");
+        }
+        return result.records;
+    } else {
+        logger.error(result.error, "Cannot parse Boomi database dump");
+        throw Error(
+            "Cannot parse Boomi database dump: " + JSON.stringify(result.error),
+        );
+    }
 }
 
 /**
--- a/backend/src/hmc-api/parser.ts
+++ b/backend/src/hmc-api/parser.ts
@@ -137,11 +137,15 @@ export type BoomiWarning =
           tag: BoomiWarningTag.ExtraneousEntryValues;
           fields_desired: number;
           fields_given: number;
+          fields: string[];
+          entryIndex: number;
       }
     | {
           tag: BoomiWarningTag.InsufficientEntryValues;
           fields_desired: number;
           fields_given: number;
+          fields: string[];
+          entryIndex: number;
       }
     | {
           tag: BoomiWarningTag.MissingSentinel;
@@ -167,17 +171,17 @@ export type BoomiResult<K extends string> =
  * @param fields A list containing the name of each field that the columns in the
  *  database correspond to. A null value means that the column at that index should be skipped.
  *  SAFETY: This list must contain at least one of every possible
- *  non-null value of the type  Fields[number].
+ *  value of the `K` type.
  * @param data A string containing a Boomi database dump.
  * @returns If parsing was successful, a result containing the list of all valid
  *  database entries parsed into records according to the fields parameter,
  *  and a list of warnings. If unsuccessful, returns the fatal error.
  */
-export function parseBoomi<K extends string, Fields extends (K | null)[]>(
-    fields: Fields,
+export function parseBoomi<K extends string>(
+    fields: (K | null)[],
     data: string,
-): BoomiResult<NonNullable<Fields[number]>> {
-    type EntryRecord = Record<NonNullable<Fields[number]>, string>;
+): BoomiResult<K> {
+    type EntryRecord = Record<K, string>;
 
     const outStart = data.match(/OUT_START\|\d+\|\@\|/);
     if (outStart === null) {
@@ -191,7 +195,9 @@ export function parseBoomi<K extends string, Fields extends (K | null)[]>(
     const boomiEntries = data.slice(dataStartIndex).split("|#|");
 
     let hasSentinel = false;
-    for (const boomiEntry of boomiEntries) {
+    for (let entryIndex = 0; entryIndex < boomiEntries.length; ++entryIndex) {
+        const boomiEntry = boomiEntries[entryIndex]!;
+
         // Empty entry is used as sentinel.
         if (boomiEntry.length === 0) {
             hasSentinel = true;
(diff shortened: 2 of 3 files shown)


backend/src/hmc-api/files.ts
```diff
@@ -96,6 +96,30 @@ export function parseCourse(data: string) {
...
+        for (const warning of result.warnings) {
+            logger.warn(warning, "Warning parsing Boomi database dump");
+        }
+        return result.records;
+    } else {
+        logger.error(result.error, "Cannot parse Boomi database dump");
+        throw Error(
+            "Cannot parse Boomi database dump: " + JSON.stringify(result.error),
+        );
+    }
```
kwshi: minor, but this feels like an appropriate occasion for a "happy path" style--
```typescript
if (!result.ok) {
  logger.error(...);
  throw Error(...);
}

for (const warning ...) {
  ...
}
```

backend/src/hmc-api/parser.ts
```diff
@@ -125,3 +125,125 @@ export function parse<Fields extends Record<string, string>>(
...
+ *  and a list of warnings. If unsuccessful, returns the fatal error.
+ */
+export function parseBoomi<K extends string, Fields extends (K | null)[]>(
+    fields: Fields,
+    data: string,
+): BoomiResult<NonNullable<Fields[number]>> {
+    type EntryRecord = Record<NonNullable<Fields[number]>, string>;
+
+    const out_start = data.match(/OUT_START\|\d+\|\@\|/);
+    if (out_start === null) {
```
kwshi: use `camelCase` (and elsewhere in this function)

backend/src/hmc-api/parser.ts
```diff
@@ -125,3 +125,125 @@ export function parse<Fields extends Record<string, string>>(
...
+ *  database entries parsed into records according to the fields parameter,
+ *  and a list of warnings. If unsuccessful, returns the fatal error.
+ */
+export function parseBoomi<K extends string, Fields extends (K | null)[]>(
+    fields: Fields,
+    data: string,
+): BoomiResult<NonNullable<Fields[number]>> {
+    type EntryRecord = Record<NonNullable<Fields[number]>, string>;
+
+    const out_start = data.match(/OUT_START\|\d+\|\@\|/);
```
kwshi: I personally think it would be nice if this were slightly more strict-- say, include the `|@|` before the `OUT_START` in the regex pattern:

```typescript
data.match(/\|@\|OUT_START\|\d\|\@\|/);
```

(by the way, do we know what the meaning of the `\d` term in the db dump is?)

backend/src/hmc-api/parser.ts
```diff
@@ -125,3 +125,125 @@ export function parse<Fields extends Record<string, string>>(
...
+    const data_start_index = out_start.index! + out_start[0].length;
+    const boomi_entries = data.slice(data_start_index).split("|#|");
+
+    let has_sentinel = false;
+    for (const boomi_entry of boomi_entries) {
+        // Empty entry is used as sentinel.
+        if (boomi_entry.length === 0) {
+            has_sentinel = true;
+            break;
+        }
```
kwshi: what does "sentinel" mean? why does there need to be a sentinel?

backend/src/hmc-api/parser.ts
```diff
@@ -125,3 +125,125 @@ export function parse<Fields extends Record<string, string>>(
...
+    ExtraneousEntryValues = "ExtraneousEntryValues",
+    InsufficientEntryValues = "InsufficientEntryValues",
+    MissingSentinel = "MissingSentinel",
+}
+
+export type BoomiWarning =
+    | {
+          tag: BoomiWarningTag.ExtraneousEntryValues;
+          fields_desired: number;
+          fields_given: number;
```
kwshi, line 139: `camelCase`

kwshi: also I think it would be nice if the warning data here tracks not just the differing numbers, but also some error-identifying information such as index/contents of the violating row, etc.

backend/src/hmc-api/parser.ts
```diff
@@ -125,3 +125,125 @@ export function parse<Fields extends Record<string, string>>(
...
+        // Check if there are extraneous columns we must ignore.
+        // If so, process only this entry's relevant values with a warning.
+        if (boomi_entry_values.length > fields.length) {
+            warnings.push({
+                tag: BoomiWarningTag.ExtraneousEntryValues,
+                fields_desired: fields.length,
+                fields_given: boomi_entry_values.length,
+            });
+        }
+        const record_kv: { [k: string]: string } = {};
```
kwshi: not a huge fan of the name `record_kv`, maybe suggest `entry` or `entryUntyped` or something

backend/src/hmc-api/parser.ts
```diff
@@ -125,3 +125,125 @@ export function parse<Fields extends Record<string, string>>(
...
+        let field_idx = 0;
+        for (const field of fields) {
+            if (field === null) {
+                ++field_idx;
+                continue;
+            }
+            // Safety: 0 <= field_idx < fields.length <= boomi_entry_values.length.
+            record_kv[field] = boomi_entry_values[field_idx]!;
+            ++field_idx;
+        }
```
kwshi: if you're going to use the index anyway you might as well do `for (let i=0; i < fields.length; ++i)`-- which, last I heard, is somehow significantly faster than `for (const .. of)` (low-quality reason) and is also more compact/idiomatic/robust (medium-quality reason). Either that or `for (const i in fields)`.

kwshi: using C-style for loop also saves you from having to manually increment in the `continue` case, which is part of why I think it's cleaner

backend/src/hmc-api/files.ts
```diff
@@ -96,6 +96,30 @@ export function parseCourse(data: string) {
     return courses;
 }
 
+export function parseCourseBoomi(data: string): {
+    code: string;
+    title: string;
+    campus: string;
+    description: string;
+}[] {
```
kwshi: I say define and export a separate named type for this
```typescript
export interface BoomiCourse {
  code: string;
  // ...
}
```
and use that in the function signature. since it's likely a type that'll get used elsewhere as well.

kwshi: generally it's useful to export any type that is slightly non-trivial; anonymous/inline types are really (ideally imo) only for small short things.

backend/tests/hmc-course.test.ts
```diff
@@ -2,10 +2,46 @@ import { describe, test, expect } from "@jest/globals";
...
 const PARSED_SAMPLE_PATH = "src/hmc-api/sample/parsed-sample-v4.json";
 
+const BOOMI_DB_SAMPLE = `
+DBSTART|85ba3f56-d039-43dd-ad22-913c0e2b42b8|2|@|BEGIN|2|@|OUT_START|3|@|SMPL010A EX|^|Example Class|^|SAFR|^||^|SC|^|Example Description
+|#|SMPL127  EX|^|Another Example Class|^|SAFR|^|05.0201|^|PZ|^|Indeed, another example description
+|#|SMPL123  EX|^|Class With Not enough fields|^|PZ|^|this one should be skipped
+|#|SMPL124  EX|^|Class With too many fields|^|SAFR|^||^|PZ|^|the next field should be ignored|^|junk|^||^|
+|#|SMPL191  EX|^|Example Class Boogaloo|^|PAFR|^||^|PO|^|Yeehaw|#||#|OUT_END|3|@|END|2|@|DBEND|85ba3f56-d039-43dd-ad22-913c0e2b42b8|2|@|
+`;
+const BOOMI_DB_EXPECTED: ReturnType<typeof parseCourseBoomi> = [
```
kwshi: would be nice to use previous suggestion of exporting the `BoomiCourse` type and referencing that instead (more explicit/direct) instead of relying on `ReturnType` inference here.

backend/tests/hmc-course.test.ts
```diff
@@ -2,10 +2,46 @@ import { describe, test, expect } from "@jest/globals";
 import type * as APIv4 from "hyperschedule-shared/api/v4";
 
 import { linkCourseData, extractSectionTerm } from "../src/hmc-api/course";
+import { parseCourseBoomi } from "../src/hmc-api/files";
 import * as fs from "fs/promises";
 
 const PARSED_SAMPLE_PATH = "src/hmc-api/sample/parsed-sample-v4.json";
 
+const BOOMI_DB_SAMPLE = `
```
kwshi: i wonder whether you should put this under a separate file instead of pasted inline. i don't care very strongly though.

kwshi (changes requested):

backend/src/hmc-api/parser.ts
```diff
@@ -125,3 +125,125 @@ export function parse<Fields extends Record<string, string>>(
...
+ *  non-null value of the type  Fields[number].
+ * @param data A string containing a Boomi database dump.
+ * @returns If parsing was successful, a result containing the list of all valid
+ *  database entries parsed into records according to the fields parameter,
+ *  and a list of warnings. If unsuccessful, returns the fatal error.
+ */
+export function parseBoomi<K extends string, Fields extends (K | null)[]>(
+    fields: Fields,
+    data: string,
+): BoomiResult<NonNullable<Fields[number]>> {
```
kwshi: oh by the way since we established that guaranteeing soundness is too hard, I think an even simpler type suffices and works about the same?
```typescript
export function parseBoomi<K extends string>(fields: (K|null)[], data: string): BoomiResult<K>
```

backend/src/hmc-api/parser.ts
foobles: Sentinel is the name of an element that signals the end of a list, e.g. the null-byte in a C string. The database format signals it's done by having a totally empty entry, so if that's not present I warn for improperly formatted input.

backend/tests/hmc-course.test.ts
```diff
@@ -78,4 +77,47 @@ describe("src/hmc-api/course.ts", () => {
...
+        console.log(JSON.stringify(boomi_courses));
+        expect(boomi_courses).toStrictEqual(BOOMI_DB_EXPECTED);
+    });
 });
+
+const PARSED_SAMPLE_PATH = "src/hmc-api/sample/parsed-sample-v4.json";
+
+const BOOMI_DB_SAMPLE = `
+DBSTART|85ba3f56-d039-43dd-ad22-913c0e2b42b8|2|@|BEGIN|2|@|OUT_START|3|@|SMPL010A EX|^|Example Class|^|SAFR|^||^|SC|^|Example Description
+|#|SMPL127  EX|^|Another Example Class|^|SAFR|^|05.0201|^|PZ|^|Indeed, another example description
```
mia1024, line 92: can we modify the test case so that one of the test data is not newline-separated from the next one?

mia1024:
Since this is functional, I'm bypassing branch protection to speed up development process. Future revisions can be added later.

PR #192, opened 2023-02-19
New Fetcher

mia1024:
This PR introduces a new data module for a static, final snapshot of each semester and the definition for a fetcher. The functionalities of the fetcher has been manually tested by calling fetchAllForTerm() for two different semesters and inspected the saved data.

I plan to replace everything in `backend/hmc-api/sample` eventually. However this cannot be done any time soon because we need to implement loaders for the new data format. In addition, we need to request a few modification for some endpoints, and that can take possibly a few weeks to complete.


backend/src/hmc-api/fetcher/endpoints.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * @author Mia Celeste
+ * @date Feb 2022
+ *
+ * This file is untracked in git and the contents will remain to be a simple mirror
+ * of endpoints-dev. We promised the school to not disclose how exactly we are contacting
+ * their server for data as a security measurement.
```
kwshi, line 7: so this file right now is just a placeholder?

backend/src/hmc-api/fetcher/fetch.ts
```diff
@@ -0,0 +1,89 @@
...
+const logger = createLogger("hmc.fetcher");
+
+async function doFetch(endpoint: Endpoint, term: APIv4.TermIdentifier) {
+    const params = computeParams(term);
+    let link: string;
+    if (endpoint.param === null) link = endpoint.link;
+    else
+        link = `${endpoint.link}?${new URLSearchParams({
+            [endpoint.param.toUpperCase()]: params[endpoint.param],
+        }).toString()}`;
```
kwshi: purely aesthetic personal preference but any chance I could invite you to do this with a `const` ternary?

```typescript
const link = endpoint.param === null ? endpoint.link : ...
```

kwshi: (i like `const` over `let` wherever possible, even though practically there isn't much of a difference)

backend/src/hmc-api/fetcher/fetch.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Fetches every single resource simultaneously. This function should not be called in production.
+ */
+export async function fetchAllForTerm(term: APIv4.TermIdentifier) {
+    if (process.env.NODE_ENV === "production")
+        throw Error("Don't fetch all data at once in production");
+    const jobs: Promise<string>[] = [];
+    for (const endpoint of Object.values(endpoints))
+        jobs.push(fetchAndSave(endpoint, term));
```
kwshi: this is logically compact enough that maybe it's simpler to do the functional thing that I suspect @foobles would love:
```typescript
const jobs = Object.values(endpoints).map(endpoint => fetchAndSave(endpoint, term))
```

backend/src/hmc-api/fetcher/types.ts
```diff
@@ -0,0 +1,33 @@
...
+import { TermIdentifierString } from "hyperschedule-shared/api/v4";
+
+/**
+ * defines which param an endpoint use, if any
+ */
+export const Param = z.union([
+    z.literal("catalog"),
+    z.literal("session"),
+    z.literal("year"),
+]);
```
kwshi, line 12: I think this should just be
```typescript
z.enum(['catalog', 'session', 'year'])
```

backend/src/hmc-api/course.ts
```diff
@@ -492,18 +493,21 @@ function processCourseAreas(
...
-    calendarSessionSection: string;
-    course: string;
-    courseSection: string;
-    courseSectionSchedule: string;
-    permCount: string;
-    sectionInstructor: string;
-    staff: string;
-    courseAreas: string;
-}): APIv4.Section[] {
+export const CourseFiles = z.object({
```
kwshi, line 501: what's the mapping here? mapping from the name we're using to refer to the file in codebase to the literal csv file name? can we add a comment/docstring explaining that?

backend/src/hmc-api/course.ts
```diff
@@ -492,18 +493,21 @@ function processCourseAreas(
...
+    altstaff: z.string(),
```
kwshi, line 502: `altStaff` :)

backend/src/hmc-api/fetcher/types.ts
```diff
@@ -0,0 +1,33 @@
...
+export const Params = z.object({
+    catalog: z.string().regex(/UG\d{2}/),
+    session: TermIdentifierString,
+    year: z.string().regex(/^20\d{2}$/),
+});
+export type Params = z.infer<typeof Params>;
+
+const Endpoint = z.object({
+    link: z.string().url(),
+    param: Param.nullable(),
```
kwshi, line 24: can you explain how this is supposed to work, since it looks like currently all the endpoints are defined with `null` param?

backend/src/hmc-api/fetcher/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+            (term.term === APIv4.Term.fall ? term.year : term.year - 1) % 100
+        }`,
+    } satisfies Params);
+}
+
+export function computeFilePath(
+    endpoint: Endpoint,
+    term: APIv4.TermIdentifier,
+) {
+    return resolve(
```
kwshi, line 30: i think it would be a little bit better if this were `path.join` instead of `path.resolve`?

backend/src/hmc-api/fetcher/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+        DATA_DIR_PATH,
+        APIv4.stringifyTermIdentifier(term),
+        endpoint.saveAs,
+    );
+}
+
+/**
+ * loads a static file from the data directory corresponding to the endpoint
+ */
+export async function loadStatic(
```
kwshi: i don't think there's any reason this needs to be `async`

backend/src/hmc-api/fetcher/utils.ts
```diff
@@ -0,0 +1,88 @@
+import * as APIv4 from "hyperschedule-shared/api/v4";
+import { DATA_DIR_PATH } from "hyperschedule-data";
+import type { Endpoint, Endpoints } from "./types";
+import { Params } from "./types";
+import { resolve, dirname } from "path";
```
kwshi, line 5: mildly prefer common/general-purpose modules to be imported as `import * as path` to keep global namespace tidier (also below, `import * as fs`)

backend/src/hmc-api/fetcher/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+    const buf = Buffer.from(content, "utf-8");
+    // writes the content of the file using utf-8 encoding, even if the underlying data might not be
+    // properly encoded. that's why we have the fixEncoding function
+    await writeFile(path, buf, {
+        encoding: "utf-8",
+    });
+    logger.info(`File ${path} written (${(buf.length / 1024).toFixed(2)} KB)`);
+}
+
+export async function loadCourseFiles(
```
kwshi, line 69: once we have a real database up and running this code becomes irrelevant, right?

data/index.ts
```diff
@@ -0,0 +1,12 @@
...
+ * since everything in node is relative to either whichever file being executed or
+ * the working directory, both of which are unreliable
+ */
+
+import * as path from "path";
+import { fileURLToPath } from "url";
+
+export const DATA_DIR_PATH = path.dirname(
+    path.resolve(fileURLToPath(import.meta.url)),
+);
```
kwshi, line 12: what's the reason we need a `path.resolve` here? do we need it at all?

kwshi (commented):
generally looks solid, just have some questions

backend/src/hmc-api/fetcher/endpoints.ts
mia1024: Yes, and it's gonna forever be left out of version control so we don't leak links and path parameters. Also the amount of settings means it's not quite practical to be env vars.

backend/src/hmc-api/fetcher/utils.ts
mia1024: No, because we still need to use the course files for unit testing and initializing a local dev database

data/index.ts
mia1024: We need this so we can import the `DATA_DIR_PATH` from anywhere, even when we just run some script (e.g. `init-db.ts` from some arbitrary working directory.

backend/src/hmc-api/fetcher/utils.ts
mia1024: Those two functions are practically identical since we aren't dealing with anything with a leading slash

backend/src/hmc-api/fetcher/utils.ts
mia1024: I think everything from `path` should be imported by value because we want to free up the `path` variable name.

kwshi (approved):
dope

PR #204, opened 2023-11-20
added "show only non-conflicting sections" filter

NextZtepS:
__Added a new slider option in the popup Filter: show only non-conflicting sections filter__

If the option is enabled, the filtered sections will only contain the sections that do not conflict with those already added to the user's current schedule.

Implementation Summary:
- add a state ```showOnlyNonConflicting``` to the ```useStore``` hook
- add a slider in the popup Filter for the user to choose the option
- retrieve the sections that are already in the user's schedule
- filter out of the conflicting sections in ```sectionsToShow``` before they get rendered to the UI

### DIFF
--- a/frontend/src/components/course-search/CourseSearch.tsx
+++ b/frontend/src/components/course-search/CourseSearch.tsx
@@ -25,8 +25,8 @@ export default memo(function CourseSearch() {
     const searchText = useStore((store) => store.searchText);
     const searchFilters = useStore((store) => store.searchFilters);
     const areas = useCourseAreaDescription().data;
-    const hideConflictingSections = useStore(
-        (store) => store.hideConflictingSections,
+    const hideConflictingSectionsOptions = useStore(
+        (store) => store.hideConflictingSectionsOptions,
     );
 
     const activeSchedule = useActiveSchedule();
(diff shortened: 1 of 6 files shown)


frontend/src/components/course-search/CourseSearch.tsx
```diff
@@ -13,13 +13,38 @@ import CourseRow from "@components/course-search/CourseRow";
...
+    function getNonConflictingSections(selectedSections: APIv4.Section[] | undefined,
+        sections: APIv4.Section[] | undefined
+    ): APIv4.Section[] | undefined {
+        if (!sections) {
+            return undefined;
+        } else if (!selectedSections) {
+            return sections;
+        } else {
+            return sections.filter(section => {
+                return selectedSections.some(selectedSection => !sectionsConflict(selectedSection, section));
```
mia1024: Since this function only has one line, you can simply do `section => selectedSections.some(selectedSection => !sectionsConflict(selectedSection, section))`

frontend/src/components/course-search/CourseSearch.tsx
```diff
@@ -13,13 +13,38 @@ import CourseRow from "@components/course-search/CourseRow";
...
+    const sectionsLookup = useActiveSectionsLookup();
+    const selectedSections: APIv4.Section[] = activeSchedule?.sections
+        .filter((us) => us.attrs.selected)
+        .map((us) =>
+            sectionsLookup.get(APIv4.stringifySectionCodeLong(us.section)),
+        )
+        .filter((s) => s !== undefined) as APIv4.Section[];
+
+    function getNonConflictingSections(selectedSections: APIv4.Section[] | undefined,
+        sections: APIv4.Section[] | undefined
```
mia1024: You don't need the `| undefined` here because neither arguments can be `undefined`, based on what you passed into this function in line 78 (they can, however, be empty arrays).

frontend/src/components/popup/Filter.tsx
```diff
@@ -65,9 +66,21 @@ const filterSpecs: { [k in Search.FilterKey]: FilterSpec } = {
...
     const addFilter = useStore((store) => store.addSearchFilter);
+    const showOnlyNonConflicting = useStore((store) => store.showOnlyNonConflicting);
+    const setShowOnlyNonConflicting = useStore((store) => store.setShowOnlyNonConflicting);
 
     return (
         <div>
+            <div>
+                <Slider
+                    value={showOnlyNonConflicting}
+                    text="show only non-conflicting sections"
```
mia1024: Instead of "show only non-conflicting sections", it's probably better to say "hide conflicting sections" or something like that. Also, the text on the legacy frontend says "Hide courses conflicting with schedule", and we can probably just keep that.

mia1024 (changes requested):

mia1024 (changes requested):
See my comments from last commit

frontend/src/components/popup/Settings.tsx
```diff
@@ -188,6 +189,34 @@ const AppearanceSettings = memo(function AppearanceSettings() {
...
+                    hideConflictingSectionsOptions.alsoHideSectionsOfSelectedCourse
+                }
+                onToggle={() => {
+                    setHideConflictingSectionsOptions({
+                        ...hideConflictingSectionsOptions,
+                        alsoHideSectionsOfSelectedCourse:
+                            !hideConflictingSectionsOptions.alsoHideSectionsOfSelectedCourse,
+                    });
+                }}
+                text="Also hide other sections of the selected course when hiding the conflicting sections"
```
mia1024: "Always show all sections of selected courses"

frontend/src/components/popup/Settings.tsx
```diff
@@ -188,6 +189,34 @@ const AppearanceSettings = memo(function AppearanceSettings() {
...
+    const hideConflictingSectionsOptions = useStore(
+        (store) => store.hideConflictingSectionsOptions,
+    );
+    const setHideConflictingSectionsOptions = useStore(
+        (store) => store.setHideConflictingSectionsOptions,
+    );
+
+    return (
+        <div className={Css.filters}>
+            <h3 className={Css.title}>Filters</h3>
```
mia1024:
```tsx
<h3 className={Css.title}>Hide Conflicting</h3>
```

frontend/src/components/course-search/CourseSearch.tsx
```diff
@@ -5,21 +5,89 @@ import { useMeasure } from "@react-hookz/web";
...
+        } else {
+            return sections.filter((section) => {
+                if (alsoHideSectionsOfSelectedCourse) {
+                    /* @desc    Check if the section is one of the selected SECTIONS first
+                     *          in case the user intentionally added this section even though
+                     *          it conflicts with other selected sections
+                     */
+                    for (const selectedSection of selectedSections) {
+                        // a particular section is non-conflicting with itself
+                        if (section.identifier === selectedSection.identifier) {
```
mia1024:
```typescript
if (Object.is(section, selectedSection))
```

frontend/src/components/course-search/CourseSearch.tsx
```diff
@@ -5,21 +5,89 @@ import { useMeasure } from "@react-hookz/web";
...
+                     */
+                    for (const selectedSection of selectedSections) {
+                        if (
+                            section.identifier.affiliation ===
+                                selectedSection.identifier.affiliation &&
+                            section.identifier.courseNumber ===
+                                selectedSection.identifier.courseNumber &&
+                            section.identifier.department ===
+                                selectedSection.identifier.department
+                        ) {
```
mia1024:
```typescript
if (APIv4.compareCourseCode(section.identifier, selectedSection.identifier))
```

frontend/src/components/course-search/CourseSearch.tsx
mia1024: Alternatively, do `APIv4.compareSectionIdentifier(section.identifier, selectedSection.identifier)` so we are not comparing object identity, in case they are different objects.

mia1024 (approved):

PR #205, opened 2023-12-04
added option to also hide async sections

NextZtepS:
Added another option in Section Conflict that ask if the user also want to hide async sections when the an option hide the conflicting sections is selected.

Also, fixed some default behavior of the Section Conflict from previous deployment

### DIFF
--- a/frontend/src/components/course-search/CourseSearch.tsx
+++ b/frontend/src/components/course-search/CourseSearch.tsx
@@ -43,34 +43,39 @@ export default memo(function CourseSearch() {
         selectedSections: APIv4.Section[],
         conflictingSectionsOptions: ConflictingSectionsOptions,
     ): APIv4.Section[] {
-        const { hidden, skipSectionsOfSelectedCourse } =
+        const { hidden, skipSectionsOfSelectedCourse, hideAsyncSections } =
             conflictingSectionsOptions;
 
         if (!hidden) {
             return sections;
         } else {
             return sections.filter((section) => {
-                if (!skipSectionsOfSelectedCourse) {
-                    /* @desc    Check if the section is one of the selected SECTIONS first
-                     *          in case the user intentionally added this section even though
-                     *          it conflicts with other selected sections (Default)
-                     */
-                    for (const selectedSection of selectedSections) {
-                        // a particular section is non-conflicting with itself
-                        if (
-                            Object.is(
-                                section.identifier,
-                                selectedSection.identifier,
-                            )
-                        ) {
-                            return true;
+                /* @desc    Check if the section belongs to one of the selected COURSES first
+                 *          in case the user intentionally added this section even though
+                 *          it conflicts with other selected sections
+                 */
+                for (const selectedSection of selectedSections) {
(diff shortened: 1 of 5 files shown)


frontend/src/components/course-search/CourseSearch.tsx
```diff
@@ -43,34 +43,39 @@ export default memo(function CourseSearch() {
...
+                            section.identifier,
+                            selectedSection.identifier,
+                        )
+                    ) {
+                        return true;
+                    }
+                }
+
+                if (hideAsyncSections) {
+                    // An asynchronous section has an empty array of scheduled days
```
mia1024: you forgot the case that `startTime === endTime`

frontend/src/hooks/store/index.ts
```diff
@@ -136,9 +137,12 @@ const initStore: Zustand.StateCreator<Store> = (set, get) => {
         setScheduleRenderingOptions: (options) =>
             set({ scheduleRenderingOptions: options }),
 
+        // TODO: add hide async classes
+        // checking for async -> start-time == end-time || no day (empty array)
```
mia1024: remove those

frontend/src/hooks/store/index.ts
```diff
@@ -136,9 +137,12 @@ const initStore: Zustand.StateCreator<Store> = (set, get) => {
...
         conflictingSectionsOptions: {
             hidden: false,
-            skipSectionsOfSelectedCourse: true,
+            skipSectionsOfSelectedCourse: false,
+            hideAsyncSections: false,
```
mia1024, line 146: we want to persist these two options (as they are set in the settings page) but not the master switch of hide conflicting (as it is a special filter from the user's perspective and filters are not saved). So you probably want to split out the `hidden` attribute and add the remaining `conflictingSectionsOptions` to the list of persisted states at the bottom of this file.

NextZtepS:
The suggestions have been implemented!

frontend/src/components/course-search/CourseSearch.tsx
```diff
@@ -43,53 +46,55 @@ export default memo(function CourseSearch() {
...
+            /* @desc    Check if the section belongs to one of the selected COURSES first
+             *          in case the user intentionally added this section even though
+             *          it conflicts with other selected sections
+             */
+            for (const selectedSection of selectedSections) {
+                // a particular section is non-conflicting with itself
+                if (Object.is(section.identifier, selectedSection.identifier)) {
+                    return true;
+                }
+            }
```
mia1024, line 62: We can merge this for now. For future reference, we can probably change this to something like 
```ts
if (selectedSection.includes(section)) return true
```
since we are only comparing object identities anyway. This can remove the for-loop, but I'm a bit wary of the correctness of this way

mia1024 (approved):

frontend/src/components/course-search/CourseSearch.tsx
NextZtepS: I will check on the correctness of that, and switch to it in the next merge if it works correctly.

PR #208, opened 2023-12-22
Add logging

stuxf:
This pull request adds logging functionality to the backend. It includes a new `discordTransport.mjs` file that provides a transport for logging to a Discord webhook. The `logger.ts` file has been updated to include the new transport when the `DISCORD_WEBHOOK_URL` environment variable is set.

### DIFF
--- a/backend/src/logger.ts
+++ b/backend/src/logger.ts
@@ -62,6 +62,18 @@ switch (env) {
                         destination: 1, // STDOUT
                     },
                 },
+                ...(process.env.DISCORD_WEBHOOK_URL
+                    ? [
+                          {
+                              target: "./discordTransport.mjs",
+                              level: "error",
+                              options: {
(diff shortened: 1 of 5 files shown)


mia1024 (changes requested):
We don't want to add the unnecessary dependency on `discord-pino-transport`. Due to the trivial nature of webhooks such functions should be written directly as a part of our code (probably a single call to `fetch` is enough).

backend/src/logger.ts
```diff
@@ -62,6 +62,18 @@ switch (env) {
                         destination: 1, // STDOUT
                     },
                 },
+                ...(process.env.DISCORD_WEBHOOK_URL
```
mia1024, line 65: There should be error checking here. If this environment variable is not set, this will be undefined or empty string and cause unexpected error. We should check that this variable is set and terminate the program startup otherwise

stuxf:
i'll work on this later lol

PR #210, opened 2024-01-03
Feature historical search

NextZtepS:
__Added a new experimental feature to search for matching sections when no section is the current term is found!__

What is added?
- new fetch: `getSectionsForTerms` -> implement on top of the existing fetch
- new hook: `useSectionsForTermsQuery` -> query the data from multiple terms only when necessary
- new sub-component: `HistoricalSearchResults` -> added in the `CourseSearch` component
- new data member: `matchingHistoricalSections` -> added in the `CourseSearch` component
- new sub-component in `Settings` popup: `HistoricalSearchSettings` -> could be used to controls the behavior of added feature

What is modified?
- conditional rendering in `CourseSearch` component for displaying the results
- `CourseRow` component now takes one more argument to specify if it belongs to other terms (could be reversed for more intuitive logic). Thus, `CourseRow` will show `SectionDetails` on click if it doesn't belong to the current term.
- some CSS property to facilitate the new `HistoricalSearchResults`
- `SectionDetails` accepts `Section` instead of `Section.identifier`

### DIFF
--- a/frontend/src/components/common/Dropdown.tsx
+++ b/frontend/src/components/common/Dropdown.tsx
@@ -1,10 +1,10 @@
 import Css from "./Dropdown.module.css";
 import * as Feather from "react-feather";
 export default function Dropdown(props: {
-    choices: (string | number)[];
-    selected: string | number;
+    choices: string[];
+    selected: string;
     onSelect: (selected: number) => void;
-    emptyPlaceholder: string | number;
+    emptyPlaceholder: string;
 }) {
     return (
         <div className={Css.dropdown}>
--- a/frontend/src/components/common/Dropdown.tsx
+++ b/frontend/src/components/common/Dropdown.tsx
@@ -1,7 +1,7 @@
 import Css from "./Dropdown.module.css";
 import * as Feather from "react-feather";
 export default function Dropdown(props: {
-    choices: string[] | number[];
+    choices: (string | number)[];
     selected: string | number;
     onSelect: (selected: number) => void;
     emptyPlaceholder: string | number;
--- a/frontend/src/components/common/Dropdown.tsx
+++ b/frontend/src/components/common/Dropdown.tsx
@@ -1,10 +1,10 @@
 import Css from "./Dropdown.module.css";
 import * as Feather from "react-feather";
 export default function Dropdown(props: {
-    choices: string[];
-    selected: string;
+    choices: string[] | number[];
+    selected: string | number;
     onSelect: (selected: number) => void;
-    emptyPlaceholder: string;
+    emptyPlaceholder: string | number;
 }) {
     return (
         <div className={Css.dropdown}>
--- a/frontend/src/components/course-search/CourseRow.tsx
+++ b/frontend/src/components/course-search/CourseRow.tsx
@@ -49,7 +49,7 @@ export default memo(function CourseRow(props: {
                     )}
                 >
                     <div className={Css.titlebar} onClick={props.onClick}>
-                        <div />
+                        <Feather.Archive className={Css.archive} size={14} />
                         <span className={Css.summary}>
                             <span className={Css.courseNumber}>
                                 {APIv4.stringifySectionCode(
--- a/frontend/src/components/course-search/CourseRow.tsx
+++ b/frontend/src/components/course-search/CourseRow.tsx
@@ -31,6 +31,7 @@ export default memo(function CourseRow(props: {
     const [detailsBounds, detailsRef] = useMeasure<HTMLDivElement>();
     const height = detailsBounds?.height ?? 0;
     const theme = useStore((store) => store.theme);
+    const setHoverSection = useStore((store) => store.setHoverSection);
 
     useEffect(() => {
         if (!props.updateDetailsSize || !detailsBounds || !props.expand) return;
@@ -82,6 +83,12 @@ export default memo(function CourseRow(props: {
                     theme,
                     false,
                 )}
+                onPointerEnter={() => {
+                    setHoverSection(props.section.identifier);
+                }}
+                onPointerLeave={() => {
+                    setHoverSection(null);
+                }}
             >
                 <div className={Css.titlebar} onClick={props.onClick}>
                     <Feather.ChevronRight className={Css.arrow} size={14} />
--- a/frontend/src/components/course-search/CourseRow.tsx
+++ b/frontend/src/components/course-search/CourseRow.tsx
@@ -41,9 +41,7 @@ export default memo(function CourseRow(props: {
         return (
             <div className={Css.padder}>
                 <div
-                    className={classNames(Css.box, {
-                        [Css.expand!]: props.expand,
-                    })}
+                    className={classNames(Css.box)}
                     style={sectionColorStyle(
                         props.section.identifier,
                         theme,
--- a/frontend/src/components/course-search/CourseRow.tsx
+++ b/frontend/src/components/course-search/CourseRow.tsx
@@ -49,10 +49,7 @@ export default memo(function CourseRow(props: {
                     )}
                 >
                     <div className={Css.titlebar} onClick={props.onClick}>
-                        <Feather.ChevronsRight
-                            className={Css.arrow}
-                            size={14}
-                        />
+                        <div />
                         <span className={Css.summary}>
                             <span className={Css.courseNumber}>
                                 {APIv4.stringifySectionCode(
--- a/frontend/src/components/course-search/CourseRow.tsx
+++ b/frontend/src/components/course-search/CourseRow.tsx
@@ -24,6 +24,7 @@ import { memo } from "react";
 export default memo(function CourseRow(props: {
     section: APIv4.Section;
     expand: boolean;
+    fromOtherTerm?: boolean;
     onClick?: () => void;
     updateDetailsSize?: (height: number) => void;
 }) {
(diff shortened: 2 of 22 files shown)


frontend/src/components/popup/Settings.tsx
```diff
@@ -228,6 +230,69 @@ const SectionConflictSettings = memo(function SectionConflictSettings() {
...
+                    <span className={Css.warning}>
+                        {" "}
+                        (this may affect performance!)
+                    </span>
+                </span>
+                <Dropdown
+                    choices={rangeOptions}
+                    selected={
+                        experimentalFeaturesOptions.historicalSearchRange !==
+                        Infinity
```
mia1024: You should probably use `null` or `undefined` instead of `Infinity` because it can cause some very slippery bug if you accidentally did arithmetic on it (e.g. `Infinity + 1 === Infinity`). If you use `null` the type checker would catch it if you accidentally did math on it.

frontend/src/components/popup/Settings.tsx
```diff
@@ -228,6 +230,69 @@ const SectionConflictSettings = memo(function SectionConflictSettings() {
...
+                            range = Infinity;
+                        } else {
+                            range = rangeOptions[index] as number;
+                        }
+                        setExperimentalFeaturesOptions({
+                            ...experimentalFeaturesOptions,
+                            historicalSearchRange: range,
+                        });
+                    }}
+                    emptyPlaceholder={NaN}
```
mia1024: Same as above but `NaN` is even worse. The `emptyPlaceholder` here is a message and should be a string to tell user nothing is selected.

frontend/src/components/common/Dropdown.tsx
```diff
@@ -1,10 +1,10 @@
 import Css from "./Dropdown.module.css";
 import * as Feather from "react-feather";
 export default function Dropdown(props: {
-    choices: string[];
-    selected: string;
+    choices: (string | number)[];
+    selected: string | number;
     onSelect: (selected: number) => void;
-    emptyPlaceholder: string;
+    emptyPlaceholder: string | number;
```
mia1024: These change are not necessary because you can simply cast everything to string (using `Number.prototype.toString`) before passing them into this component

frontend/src/components/course-search/CourseRow.tsx
```diff
@@ -36,10 +37,51 @@ export default memo(function CourseRow(props: {
         props.updateDetailsSize(detailsBounds.height);
     }, [detailsBounds?.height, props.updateDetailsSize, props.expand]);
 
+    if (props.fromOtherTerm) {
+        return (
+            <div className={Css.padder}>
+                <div
+                    className={classNames(Css.box, {
+                        [Css.expand!]: props.expand,
```
mia1024: you don't need the non-null assertion here

frontend/src/components/course-search/CourseRow.tsx
```diff
@@ -36,10 +37,51 @@ export default memo(function CourseRow(props: {
         props.updateDetailsSize(detailsBounds.height);
     }, [detailsBounds?.height, props.updateDetailsSize, props.expand]);
 
+    if (props.fromOtherTerm) {
```
mia1024, line 41: isn't this whole block essentially the same as the thing beneath it? why the code duplication?

frontend/src/components/course-search/CourseSearch.tsx
```diff
@@ -87,25 +92,74 @@ export default memo(function CourseSearch() {
...
+    );
+    const allTerms = useAllTerms() ?? [];
+    const range = Math.min(allTerms.length, historicalSearchRange);
+    const historicalSections = useSectionsForTermsQuery(
+        enableHistoricalSearch && sectionsToShow?.length === 0,
+        allTerms.slice(0, range),
+    ).data;
+
+    const matchingHistoricalSections: APIv4.Section[] | undefined =
+        React.useMemo(() => {
```
mia1024: you are performing the entire historical search here whether the result is needed or not and is probably causing the performance issue

frontend/src/components/course-search/CourseSearch.tsx
```diff
@@ -87,25 +92,74 @@ export default memo(function CourseSearch() {
...
+                    <CourseSearchEnd text="loading courses..." />
+                )}
+
+                {enableHistoricalSearch && sectionsToShow?.length === 0 ? (
+                    <HistoricalSearchResults
+                        sections={matchingHistoricalSections}
+                    />
+                ) : (
+                    <></>
+                )}
```
mia1024: This should be modified to only perform the historical search if the matched section count is 0 (and it's probably better to make it less than 5 or something like that). Currently your code only renders the result if there is no matched section, but performs the search unconditionally. Compared to the cost of performing the searches, cost of rendering is almost negligible (on a single term the time to render everything is about 3ms on my computer but the time to perform the search is sometimes 20-30ms).

frontend/src/components/popup/Settings.tsx
NextZtepS: I used `Infinity` as an encoding for the case where the user want to search through "all" historical terms. I exclusively used this it with comparison such that we would keep the `min` between the user's desired range (2, 4, ..., 10, `Infinity`) and how many terms we have because this will ensure that we would never look back more than the terms we have in the database, and keep the case where the user want to search through all terms simply represented `infinity` range. If this causes problem, I could try encode the "all" case the other way.

frontend/src/components/common/Dropdown.tsx
NextZtepS: I will revert it back, and use string representation as you recommended.

frontend/src/components/course-search/CourseRow.tsx
NextZtepS: I will fix it.

frontend/src/components/course-search/CourseRow.tsx
NextZtepS: My idea was to separate the case where the `CourseRow` component represents the course from the non-current term out and display a much simpler version of `CourseRow` since it would not make sense for the `CourseRow` component to be expandable if it doesn't belong to this term (also, would affect course description map query to be much harder to handle). To achieve that, I wanted it serve as a anchor tag to their respective term for simplicity, so I use this if-check to render a more simple `CourseRow` element which is different from the existing `CourseRow`. 

I agree that there are quite a few duplications which could be cleaned up. I only worried that it would be overly complicated than having 2 cases separately like I did since the "foreign-term" `CourseRow` element are much simpler. (discussion needed)

frontend/src/components/popup/Settings.tsx
mia1024: You don't need to guess how many terms we have in the database. You can get a list of all of them with `useAllTerms()` in `hooks/terms.ts`

frontend/src/components/course-search/CourseSearch.tsx
NextZtepS: I will fix this by implement if-check that prevent the "expensive" course searching algorithm by returning `undefined` when the data is not needed (done easily) --> this should remove any unnecessary computation down to a simple check while we keep using the useMemo to get the benefit of simple memoization.

I will also look into the hooks you recommend, and see if they yield better performance.

frontend/src/components/course-search/CourseSearch.tsx
NextZtepS: I agree to have the historicalSearching done if the matching sections from current terms are less than some threshold (5 sounds good), and the performance issue will be handled like comment above.

frontend/src/components/popup/Settings.tsx
NextZtepS: I will fix this part to make it makes more sense!

frontend/src/components/popup/Settings.tsx
mia1024: I guess what I'm saying is, the list of options should be dynamically computed based on the number of past semesters available instead of a hard-coded list of 2,4...10, all. For example, the choice can be all the powers of 2 underneath the max (including 1), plus the max itself as a number.

frontend/src/components/popup/SectionDetails.tsx
```diff
@@ -1,29 +1,56 @@
...
+//                     activeTerm.year === props.section.year
+//                     ? APIv4.stringifySectionCode
+//                     : APIv4.stringifySectionCodeLong)(props.section)}
+//                 : {s.course.title}
+//             </h3>
+
+//             <CourseDescriptionBox section={s} showStatus={true} />
+//         </div>
+//     );
+// });
```
mia1024: don't comment out big blocks of code like this. if we don't need it anymore, delete it. if it turned out we need to revert to an earlier version we can get it from git

frontend/src/components/course-search/CourseSearch.tsx
```diff
@@ -278,3 +337,59 @@ const CourseSearchRow = memo(function CourseSearchRow(props: {
...
+            <div className={Css.emptyHistoricalSearchResults}>
+                (no historical records of courses found)
+            </div>
+        );
+    }
+
+    return (
+        <div className={Css.historicalSearchResults}>
+            <h4>Are you looking for these sections from recent terms?</h4>
+            <div>Click on any section to see its details!</div>
```
mia1024: We don't need to say things like this. For every section ever listed anywhere, clicking on the section always shows details. So this is not a special behavior and we shouldn't need to point it out like this. Also, instead of "Are you looking for these sections from recent terms?" we can just say "found results from previous terms" or something similar.

NextZtepS:
I wonder if we should merge this to v2, so that people can test it out on nightly. @mia1024

mia1024:
sure

PR #212, opened 2024-02-22
Added hoverSection hook, applied to schedule ui

edonson2016:


mia1024 (approved):
Looks good. Considering what happened to the first commit please choose squash and merge so you don't introduce broken commit into the history. The next step would be highlighting the sections in `SelectedList` as well when it's hovered over. Also, when hovering over `SelectedList`, you set the hover section without first checking if the section is selected (`section.attrs.selected`). This is technically incorrect as we shouldn't highlight anything that's not selected. This happens to be fine this time because unselected sections are not rendered, but could be a source of future bugs. You should also fix this in the future.

PR #215, opened 2024-03-21
add a span to work around oversized dropdown on safari

NextZtepS:
fix the the buggy Dropdown rendering on safari by putting in a <span></span>


mia1024 (approved):
classic safari moment

PR #217, opened 2024-04-01
Set Background Color on Dropdowns

CoolSpy3:
Currently, the background color on dropdown menus is unset. This leads to the background occasionally being white in dark-mode (ex. on the term selection dropdown), making the text in the dropdown hard to read. This PR explicitly sets the background color to the background color variable, fixing this issue.

### DIFF
--- a/frontend/src/components/common/Dropdown.module.css
+++ b/frontend/src/components/common/Dropdown.module.css
@@ -20,6 +20,7 @@
 
         background-color: transparent;
         color: var(--fg-0);
+        background-color: var(--bg-0);
         transition: color var(--misc-transition-duration);
 
         & option:disabled {
--- a/frontend/src/components/common/Dropdown.module.css
+++ b/frontend/src/components/common/Dropdown.module.css
@@ -18,7 +18,6 @@
         width: 100%;
         height: 100%;
 
-        background-color: transparent;
         color: var(--fg-0);
         background-color: var(--bg-0);
         transition: color var(--misc-transition-duration);


mia1024:
Just curious, what OS + browser are you using? We have previously encountered this issue but it seemed to be only reproducible on Firefox on certain Linux distros

stuxf:
image
Bug still seems present for me on chrome + fedora (running gnome), did you test with more than one item in the menu? It seems like the transparent css is overriding the set background

CoolSpy3:
@mia1024 I'm currently running Brave v1.64.113 on Windows 10 (64-bit). I can also reproduce the issue on Chrome v123.0.6312.86 and Firefox 124.0.1.

@stuxf I tested these changes locally and they work for me, although I had some trouble with the database, so I had to add the additional select options manually.

CoolSpy3:
I see what you mean about the transparent tag. I missed that. I'll remove it.

mia1024:
I remember there was a reason why I added the transparent background. I have a faint recollection that it might have something to do with how mobile safari renders it. But having the background color set to bg-0 is probably the more correct behavior.

stuxf (approved):
looks good, seems to fix issue for linux systems

mia1024:
@CoolSpy3 can you please verify this works at nightly.hyperschedule.io after about 10 minutes (github cache duration)

@stuxf let's not deploy this for the time being so we can test this change actually works across all browsers, especially variants of safari

CoolSpy3:
Confirmed! The bug is fixed for me on all browsers I mentioned above!

mia1024:
@CoolSpy3 we have verified and deployed the change to production site

PR #219, opened 2024-04-20
Move Theme Slider

CoolSpy3:
At the moment, the theme slider sits on top of the currently selected courses list. This makes it impossible to remove the last course in the list without first shifting it up 1 position so the delete button is exposed.

This PR moves the slider to the top of the sidebar panel so that the slider does not obstruct any GUI elements.

### DIFF
--- a/frontend/src/components/SelectedList.module.css
+++ b/frontend/src/components/SelectedList.module.css
@@ -12,6 +12,7 @@
         display: flex;
         width: 100%;
         justify-content: space-evenly;
+        height: 3rem;
     }
 
     &.isDragging {
--- a/frontend/src/components/SelectedList.tsx
+++ b/frontend/src/components/SelectedList.tsx
@@ -194,7 +194,7 @@ export default memo(function SelectedList() {
                     </DndCore.DndContext>
                     {scheduleRenderingOptions.showConflicting ||
                     activeSchedule.sections.length === 0 ? (
-                        <></>
+                        <div className={Css.textContainer}></div>
                     ) : (
                         <div className={Css.textContainer}>
                             <span>(drag sections to change priorities)</span>


stuxf:
Hey, thanks for bringing up this issue.

Interestingly enough, this only happens when show conflicting sections is selected. 

image

When show conflicting sections is not selected, the delete button is still reachable, but it's difficult and definitely still bad design. Note that it shows the "(drag sections to change priorities)" box at the bottom.

image

Perhaps a better solution to this would be to introduce a bounding box at the bottom?, so that you can't scroll down as much, or to permanently fix a box there thats height at least covers the theme slider.  Alternatively, we can try making the "drag to scroll" thing always visible and at the same horizontal level as the theme slider so there isn't a weirdly empty space.

Moving the theme slider to the top of the sidebar does feel a little bit out of place and it would be better for it to stay at the bottom right corner. Thoughts on doing something such as above (other ideas also welcome) instead?

CoolSpy3:
How about this? I increased the size of the "Drag to scroll" box and made it always present. I was originally concerned that it would look weird due to the extra space, but because you have to scroll down for it to become visible, I think it makes sense.

NextZtepS:
I support the idea of making the SelectedList element has a padding on the bottom such that the last selected course will not be overlap with the theme slider when the user scroll to the end of the list. This way, I think the problem could be solved whether "show conflicting sections" is enabled or not.

It would look something like this

Screenshot 2024-04-20 at 5 54 46 PM

if anyone thinks this might look weird, we could work on the @CoolSpy3's idea to make the "(drag sections to change priorities)" box always possible and make it a bit bigger. If we decide to move on with that, I think we could be better off adding the padding-bottom rather than increasing the size of "(drag sections to change priorities)" box.

CoolSpy3:
I should clarify: I made the box always visible, but it still only shows the text if "Show Conflicting Sections" is disabled. i.e. It looks like the image you posted.

mia1024:
I'm on the clinic site visit right now and am unable to review the PR. However, I will add that the reason why the "drag to reorder" text only shows up when show conflicting is disabled is because the ordering is currently only used for conflict resolution, which is not necessary if all the sections are shown. However, there might be some other reason why we want to have the text to be always present that I didn't think of.

stuxf (approved):
seems good to me, fairly standard change as suggested tyty

PR #220, opened 2024-04-21
Fix Pre-Commit Hook When Auto-CRLF is Enabled

CoolSpy3:
Prettier expects files to have `lf` line endings, but when the pre-commit hook runs `git stash`, all the `lf` endings are replaced with `crlf` if `autocrlf` is enabled. As a result, the pre-commit hook fails even after running prettier. This PR forces `autocrlf=lf`, disabling this behavior.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-* text eol=lf
+* text=auto eol=lf


CoolSpy3:
Updated to follow [Prettier's .gitattributes recommendation](https://prettier.io/docs/en/options.html#end-of-line)

mia1024:
What system are you using? Hyperschedule server is not really designed to be ran on Windows natively without docker or WSL, if that's what produced the CRLF ending.

CoolSpy3:
I'm developing on Windows 10 with Docker. The CRLF ending is a result of Git. Because Windows is weird, it's common practice to configure git to push to GitHub with LF line endings, but to use CRLF endings in the working tree[[1](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf)]. As a result, when the pre-commit hook runs `git stash`, git replaces some of the LF line endings with CRLF, so `prettier-check` fails.

Adding the `.gitattributes` file tells git to preserve LF line endings in the working tree even if `autocrlf` is enabled globally. [[2](https://prettier.io/docs/en/options.html#end-of-line)]

stuxf (approved):
Seems like there are not any downsides to merging this, looks good to me.

PR #221, opened 2024-08-08
updateUser function and testing a way to lowercase eppn in the db

NextZtepS:
I have write the updateUser function that could be used to make the eppn in the db to be lowercase.

The last thing that we should do it runs the scripts that follow the added test in the db to finish off the casing problem with the CAS.

### DIFF
--- a/backend/src/db/models/user.ts
+++ b/backend/src/db/models/user.ts
@@ -467,7 +467,7 @@ export async function findDuplicatesWith<K extends keyof APIv4.ServerUser>(
             },
             {
                 $match: {
-                    count: { $gt: 1 }, // Only groups with more than 1 document
+                    count: { $gt: 1 },
                 },
             },
             {
@@ -480,4 +480,14 @@ export async function findDuplicatesWith<K extends keyof APIv4.ServerUser>(
         .toArray();
 
     return result.map((group) => group.users);
+}
+
+export async function makeAllEPPNLowercase(): Promise<void> {
+    await collections.users.updateMany({ eppn: { $regex: /[A-Z]/ } }, [
+        {
+            $set: {
+                eppn: { $toLower: "$eppn" },
+            },
+        },
+    ]);
 }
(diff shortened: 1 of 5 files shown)


backend/tests/db/user.test.ts
```diff
@@ -288,4 +289,30 @@ describe("db/models/user", () => {
             selected: false,
         } satisfies APIv4.UserSectionAttrs);
     });
+
+    test("replace capital letters with lowercase in saml", async () => {
+        const uid1 = await getOrCreateUser("First Test User", "");
+        const uid2 = await getOrCreateUser("Second Test User", "");
+        const uid3 = await getOrCreateUser("third test user", "");
+        const issuerId = await getOrCreateUser(
+            "JBao00@cmc.edu",
```
mia1024: Try not to use the name of a real student for test data

backend/src/db/models/user.ts
```diff
@@ -84,6 +84,31 @@ export async function getUser(userId: string): Promise<APIv4.ServerUser> {
...
+    const lookup = await collections.users.findOne({ _id: userId });
+
+    if (lookup === null) {
+        throw new Error(`User with ID ${userId} not found`);
+    }
+
+    if (Object.keys(updateFields).length > 0) {
+        await collections.users.updateOne(
+            { _id: userId },
+            { $set: updateFields },
```
mia1024, line 100: We probably should sanitize the data at least a little bit before throwing it into the database. Otherwise we might accidentally have a data corruption down the line.

For example, if `updateFields` happen to have a method defined as part of the object (perfectly fine for a `Record<string, any>`), then this will attempt to write a function object into the database, which is obviously bad.

backend/tests/db/user.test.ts
NextZtepS: Got it. I adjusted the test case.

backend/src/db/models/user.ts
NextZtepS: I update the argument to be of type `Partial<APIv4.ServerUser>` which should only allow valid user partial to be taken as the `updateFields`. Do you see any some uncommon object type that could slip through this TS argument safeguard?

PR #224, opened 2024-09-15
Add Filter for Section Status

ohowe1:
Unsure about how to do open/reopened. I was thinking of combining them to "(Re)Open" as for what I know it doesn't matter for registration if it is open or reopened. Not sure about the "(Re)Open" title.

### DIFF
--- a/frontend/src/components/course-search/filter-bubble/StatusBubble.tsx
+++ b/frontend/src/components/course-search/filter-bubble/StatusBubble.tsx
@@ -1,0 +1,28 @@
+import * as Search from "@lib/search";
+import type { FilterBubbleComponentProps } from "./FilterBubble";
+import AutoComplete from "./AutoComplete";
+import { memo } from "react";
+
+export default memo(function StatusBubble(
+    props: FilterBubbleComponentProps<Search.StatusFilter>,
+) {
+
+    const statusOptions = Object.values(Search.StatusFilterOptions);
+
+    return (
+        <AutoComplete
+            onSelect={(index) => {
+                const status = statusOptions[index];
+                if (status !== undefined) {
+                    props.onChange({ status });
+                    props.focusNext();
+                }
+            }}
+            onKeyDown={props.onKeyDown}
+            choices={statusOptions}
+            render={({ index }) => {
+                return <span>{statusOptions[index]}</span>;
+            }}
+        />
+    );
+});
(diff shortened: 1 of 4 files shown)


mia1024:
Haven't looked at your code yet, but it does matter whether a section is reopened. All reopened sections automatically become perm-only, in addition to all the requisites they have

ohowe1:
Didn't know that. Thanks. Should be fixed

frontend/src/lib/search.ts
```diff
@@ -353,6 +359,9 @@ export function filterSection(
                     section.identifier.half.number.toString()
                 )
                     return false;
+                break;
+            case FilterKey.Status:
+                return section.status === filter.data.status;
```
stuxf: There's an early return here, you should only return false if the filter doesn't match. This is so the for loop can keep iterating through subsequent filters.

You'll notice that this breaks behavior as any filters added after a status filter will not work.

frontend/src/components/course-search/filter-bubble/StatusBubble.tsx
```diff
@@ -0,0 +1,36 @@
+import type * as Search from "@lib/search";
+import type { FilterBubbleComponentProps } from "./FilterBubble";
+import * as APIv4 from "hyperschedule-shared/api/v4";
+import AutoComplete from "./AutoComplete";
+import { memo } from "react";
+
+const StringToSectionStatusMap: Record<string, APIv4.SectionStatus> = {
```
stuxf, line 7: This includes "Unknown" on the frontend for the filter, which feels unnecessary. Was wondering why you chose to include it. I'm okay with this, just unsure about the reasoning.

stuxf (changes requested):
Thank you for your PR and patience during the review process! 

There's just one small issue with your PR, it prevents all filters applied after it from being applied. I've pointed you to where it occurs in my review.

This is how filtering should look like:

image

But notice it breaks the filters after it.

image

Let me know if you need any help implementing this or with any other questions you might have. :smile:

frontend/src/components/course-search/filter-bubble/StatusBubble.tsx
ohowe1: My thought was that it's an option in the data, so maybe someone would want to use it. Thinking about it, it seems that that use case would be pretty obscure and it might confuse some users so I think I agree that it should be removed.

frontend/src/lib/search.ts
ohowe1: Oh, missed that it was part of a loop to check all the filters. Should be fixed in the push

stuxf (approved):

stuxf:
Thank you!

PR #226, opened 2025-05-05
Add libcrypto dep to fix CI

stuxf:
Lot of larger issues here that will need to be fixed later, this is just a short term bandaid solution.

Will want to upgrade node version, and then bump the mongodbmemoryserver version, and then resolve a bunch of dependencies and typechecks that cause everything else to fail when versions are bumped


NextZtepS (approved):
I think this should be good to go to as long as the library on hosted on the legacy url. If it is not there anymore, we will tackle this again.

PR #231, opened 2025-06-22
Optimize search

stuxf:
should make search faster? currently running into a UI bottleneck where because multi term search results aren't virtualized, they take pretty long to load and can be quite laggy initially

### DIFF
--- a/base.dockerfile
+++ b/base.dockerfile
@@ -1,9 +1,4 @@
 FROM node:18-bullseye
 
-# install the linux version of all dependencies
-RUN apt update && apt install -y git
-COPY . /hyperschedule/
-
-WORKDIR /srv
-RUN git clone /hyperschedule
+COPY . /srv/hyperschedule/
 WORKDIR /srv/hyperschedule/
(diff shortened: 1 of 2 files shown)


mia1024:
BTW make sure to test the docker change on mac. The reason why that was a git clone was to exclude dependencies compiled for mac, which would cause bizarre errors when ran under a linux kernel and somehow reinstalling dependencies didn't fix it

stuxf:
wtf i don't have a mac to test on gg

ohowe1:
> wtf i don't have a mac to test on gg

works on my mac!

NextZtepS (approved):
I think it looks good! Maybe in the future. It might be wise to come up with a universal way to render both our usual `sectionsToShow` and `MultiTermsSearchResults`, so our `CourseSearch.tsx` can be concise.

stuxf:
yeah I think this could definitely be made more concise and cleaned up

stuxf:
merging for now. this shaves off most of the rendering time (~1.5 s for me w/ 4 semesters, ~30s for all semesters) but still leaves about 1s of scripting time that could be optimized. unconfident that there are no bugs with this but we can test in nightly

PR #233, opened 2025-07-10
Fix old link in README

ohowe1:


stuxf:
this one was my bad for being too lazy to update the readme w/ git lfs and telling claude to do it :clown_face:

stuxf (approved):

PR #236, opened 2025-07-29
Upgrade dependencies

stuxf:
Long overdue, this will be annoying because a lot of annoying changes.

So far:
- Update NodeJS from 18 to 22.


stuxf:
somewhere between starting and now the announcement card for frontend has had a regression oops

PR #249, opened 2026-08-26
feat: add hmc/non-hmc credits to DB to allow for overwriting typical formula where necessary

### LINKED ISSUE #179: Pomona College Orchestra is 1 credit, not 1.5
At Pomona it's 0.5, but at Mudd it's 1. (not 0.5*3).

ohowe1:
Closes #179


mia1024:
everything here looks like it should work as intended. However, my question is: why are we doing the conversion at the write-to-database step instead of doing this conversion in the frontend? Seems to me that they would both work just-as-well while saving a database migration?

ohowe1:
Hey sorry got super busy, my thought for this approach was that fixing a data issue wouldn't require a new frontend deploy and (kinda the same thing) I wanted to keep hardcoded data out of the frontend. But I can for sure see it both ways

mia1024:
Typically I think it's a good idea for there to be a source-of-truth, with all transformed data stored separately and computed on-the-fly. The basic idea is, what if we decide one day that the transformation was incorrect/unnecessary? Depending on the nature of the transformation, if we store transformed instead of the raw data, we might not be able to get the raw data back at that point. 

Now, for the particular use-case of hyperschedule, it's a little bit less clear because the db isn't really the source-of-truth; portal is. In our particular architecture, our course db is more similar to a caching service of portal than the ground truth, which means any transformation can be easily reverted by re-fetching all the data from upstream. 

As for architecture decision goes, there's actually a secret third option, which is to apply the transform at the API level. That is, when a client requests the data, first run the transformation, then send it out. The server load in that case will be minimal because of the aggressive caching we have, and that allows us to change the transformation as needed without mutating the source data or hard-code it in the front end.

ohowe1:
One thing to note is that we still have the `credits` field which stores the raw from portal credit count and that is used to make `HMCCredits` and the `nonHMCCredits`. The DB serving just as a cache of portal is interesting but I still think there's a precedent for us doing this kind of transformation when we get data from portal instead of on the API like with the building names. But that's also not quite the same thing as the credit thing here.

PR #251, opened 2026-09-11
Added class color by campus theme

### LINKED ISSUE #53: Organize course colors by college
Instead of having them completely randomized, perhaps randomize the shade by campus so it's easier to distinguish at a glance what college the class is at

InventBoss:
New color scheme in settings which will color classes across the UI by affiliation (campus). Most color choices mach up with the colors on the Claremont Colleges website, with the exception of Harvey Mudd in pink as the generated colors were hard to differentiate. Classes that have their affiliation not to a campus will be grey with the exception of the Keck Science center in green (Lot of scripps affiliated courses, thus green to match) and Joint Programs in pink.

(resolves #53)

### DIFF
--- a/frontend/src/components/course-search/CourseRow.tsx
+++ b/frontend/src/components/course-search/CourseRow.tsx
@@ -51,7 +51,7 @@ export default memo(function CourseRow(props: {
                         props.section.identifier,
                         theme,
                         courseColorTheme,
-                        props.section.identifier.affiliation,
+                        props.section.course.primaryAssociation,
                         false,
                     )}
                 >
@@ -85,7 +85,7 @@ export default memo(function CourseRow(props: {
                     props.section.identifier,
                     theme,
                     courseColorTheme,
-                    props.section.course.code.affiliation,
+                    props.section.course.primaryAssociation,
                     false,
                 )}
                 onPointerEnter={() => {
--- a/frontend/src/components/course-search/CourseRow.tsx
+++ b/frontend/src/components/course-search/CourseRow.tsx
@@ -51,6 +51,7 @@ export default memo(function CourseRow(props: {
                         props.section.identifier,
                         theme,
                         courseColorTheme,
+                        props.section.identifier.affiliation,
                         false,
                     )}
                 >
@@ -84,6 +85,7 @@ export default memo(function CourseRow(props: {
                     props.section.identifier,
                     theme,
                     courseColorTheme,
+                    props.section.course.code.affiliation,
                     false,
                 )}
                 onPointerEnter={() => {
--- a/frontend/src/components/popup/Settings.module.css
+++ b/frontend/src/components/popup/Settings.module.css
@@ -131,7 +131,7 @@
         }
 
         & .button[data-theme="campus"] {
-            --theme-color: hsla(0, 0%, 0%, 0%);
+            --theme-color: #ff6b6b;
             background: radial-gradient(circle, #feca57 0%, #ff6b6b 100%);
         }
 
--- a/frontend/src/components/popup/Settings.module.css
+++ b/frontend/src/components/popup/Settings.module.css
@@ -130,6 +130,11 @@
             --theme-color: hsl(133, 29%, 37%);
         }
 
+        & .button[data-theme="campus"] {
+            --theme-color: hsla(0, 0%, 0%, 0%);
+            background: radial-gradient(circle, #feca57 0%, #ff6b6b 100%);
+        }
+
         /* Selected state */
 
         & .button.selected {
--- a/frontend/src/lib/color.ts
+++ b/frontend/src/lib/color.ts
@@ -8,7 +8,8 @@ export type CourseColorTheme =
     | "pastel"
     | "sunset"
     | "cool"
-    | "earthy";
+    | "earthy"
+    | "campus";
 
 export const courseColorThemes: { id: CourseColorTheme; label: string }[] = [
     { id: "default", label: "Default" },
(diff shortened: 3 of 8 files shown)


frontend/src/lib/color.ts
```diff
@@ -206,7 +227,7 @@ export function sectionColorStyle(
             break;
 
         case "default":
```
ohowe1, line 227: Having a `case "campus":` here to be explicit I think would be good

frontend/src/components/course-search/CourseRow.tsx
```diff
@@ -51,6 +51,7 @@ export default memo(function CourseRow(props: {
                         props.section.identifier,
                         theme,
                         courseColorTheme,
+                        props.section.identifier.affiliation,
```
ohowe1: Honestly not very opposed to using primaryAssociation here instead of affiliation because the campus filter we have uses that:

Image

frontend/src/components/popup/Settings.module.css
```diff
@@ -130,6 +130,11 @@
             --theme-color: hsl(133, 29%, 37%);
         }
 
+        & .button[data-theme="campus"] {
+            --theme-color: hsla(0, 0%, 0%, 0%);
+            background: radial-gradient(circle, #feca57 0%, #ff6b6b 100%);
```
ohowe1, line 135: I like the gradient! One small issue is when I have it selected the border doesn't appear as it does with the others

frontend/src/lib/color.ts
```diff
@@ -8,16 +8,30 @@ export type CourseColorTheme =
...
     { id: "pastel", label: "Pastel" },
     { id: "sunset", label: "Sunset" },
     { id: "cool", label: "Cool" },
     { id: "earthy", label: "Earthy" },
+    { id: "campus", label: "Per Campus Colors" },
 ];
 
+const affiliationColors: Record<string, string> = {
+    PO: "blue", // Pomona
+    HM: "pink", // Harvey Mudd
```
ohowe1: I think the yellow doesn't look _too_ bad but I can for sure see both sides

frontend/src/lib/color.ts
```diff
@@ -8,16 +8,30 @@ export type CourseColorTheme =
...
+    PZ: "orange", // Pitzer
+    CM: "red", // Claremont McKenna
+    SC: "green", // Scripps
+    CG: "purple", // Claremont Graduate
+    KS: "green", // Keck Science Center
+    JP: "pink", // Joint Programs
+    Random: "random", // Random
```
ohowe1: Is this used for anything?

ohowe1 (commented):
Honestly really great job! Very well scoped and executed. Few notes let me know what you think

frontend/src/lib/color.ts
InventBoss: Just didn't do it because the `default:` case handled it perfectly. I understand from a legibility point though and can quickly add it in the next commit if you want. Let me know which is best!

frontend/src/components/course-search/CourseRow.tsx
InventBoss: The main problem with primary association is that `UserSection` (the data type of sections the user has selected), doesn't store `primaryAffiliation` data. I figured it'd be much easier to simply use `affiliation` and color the few sections that have weird affiliations (like Keck Science Center affiliation, which seems to mostly be used by Scripps). The rest I either assigned a color based on what colleges were most associated with them, or left them grey, in order to contrast better with the colors. From what I've observed, affiliation seems to link up with the text after the course identifier: **AF**-01, **SC**-01, etc. 

Let me know if you want me to put in more by having the user selections store `primaryAffiliation` somehow (likely not by changing `UserSection` as that data is stored on the database and I don't want to touch that right now).

frontend/src/components/popup/Settings.module.css
InventBoss: Now that I zoom in I can just barely see it! Will be followed up in the next commit.

frontend/src/lib/color.ts
InventBoss: See above decision on using `affiliation` over `primaryAffiliation`.

ohowe1: Sorry github did a bad job of showing it, this is looking at Random

frontend/src/lib/color.ts
ohowe1: Yes let's add it just for completeness

frontend/src/lib/color.ts
InventBoss: Ah I see. Looking at it now, that was a remnant of my previous approach where I'd pass random into the affiliationRecords when per campus colors was enabled. Simply old code from the previous implementation. Will get rid of in next commit.

frontend/src/components/course-search/CourseRow.tsx
ohowe1: Ah good point! What we can do though is use the `UserSection`'s section identifier to look up the real `Section` with something like `useActiveSectionsLookup`. In `SelectedList` we already do this with:
```
    const section = sectionsLookup.get(
        APIv4.stringifySectionCodeLong(props.entry.section),
    );
```

InventBoss: Thanks for the code example! Will get this implemented right away. Will make the colors much more consistent.

Thank you for your quick responses! Absolute pleasure to work on this project and I can only hope to be in your shoes one day.

frontend/src/lib/color.ts
InventBoss: Main problem for me was just differentiation. Pink was just much better (no idea why). I tried a few shades of yellow as well, but it just didn't do much.

image

image

Let me know which one you want. I'll add it onto the commit.

frontend/src/lib/color.ts
ohowe1: Oh sorry I meant more:
```
case "campus":
case "default":
default:
            color = [colorOut[0], colorOut[1], colorOut[2]];
            break;
```

frontend/src/components/course-search/CourseRow.tsx
ohowe1: Genuinely thank you so much for taking initiative, it is great to have you working on the project!

frontend/src/lib/color.ts
ohowe1: Ahh yeah its definitely tough; up to you but another thing we can do is define custom color ranges that the random color lib can pick from so that would allow us to have a wider yellow for mudd, and a more "scripps green", etc. But I am also happy with the yellow even though its a little difficult to differentiate just because its more consistent with the rest of the schools.

InventBoss: What I'm thinking rn is have it be a bespoke theme, and then I can later have it be a toggle that influences the themes. For know it'll be too difficult to have bespoke campus colors for themes like the sandy one. Lmk what you think.

ohowe1 (approved):

ohowe1:
I think we just need to run a `pnpm format` and it should be all good to merge :)

